PR #113, opened 2018-11-19
fixing BALLERINA_HOME inference issue for linux resource, fixes #112

### LINKED ISSUE #112: BALLERINA_HOME set for v0.983.0-1 on linux/rpm is wrong
**Description:**
I've installed from ballerina.io provided RPM `v0.983.0-1.x86-64`.  It sets `BALLERINA_HOME=/usr/lib64/ballerina/ballerina-0.983.0`, but the actual value as per dir structure created need to be `/usr/lib64/ballerina/ballerina-platform-0.983.0`.
This in turn results into wrong error message as well, which is `Error: JAVA_HOME is not defined correctly.`.

**Affected Product Version:**
* found it for `v0.983.0-1.x86_64`

**OS, DB, other environment details and versions:**    
* Fedora (GNU/Linux)

**Steps to reproduce:**
* Just install this version ballerina rpm on any linux and try use it.

I've fixed it locally and will submit a patch.

abhishekkr:
Signed-off-by: AbhishekKr <abhikumar163@gmail.com>

## Purpose
This fixes reported issue#112.

## Goals
It will allow linux resource to infer BALLERINA_HOME if auto-configured env var is missing or wrong.

## Approach
If auto-configured env var BALLERINA_HOME is missing or wrong, it checks for relevant dir-names under `/usr/lib64/ballerina/` and fixes the value.

## Release note
This fixes BALLERINA_HOME issue causng `Error: JAVA_HOME is not defined correctly.` for linux (at least rpm) release `v0.983.0-1`.

## Documentation
N/A. This is just a fix for already existing flow.

## Certification
N/A

## Automation tests
N/A

## Security checks
N/A

## Samples
N/A

## Related PRs
N/A

## Migrations (if applicable)
N/A

## Test environment
N/A
 
## Learning
N/A

### DIFF
--- a/resources/executables/linux/ballerina
+++ b/resources/executables/linux/ballerina
@@ -32,7 +32,7 @@
 [ -f /etc/profile.d/wso2.sh ] && source /etc/profile.d/wso2.sh
 
 if [[ ( -z "${BALLERINA_HOME}" || ! -d "${BALLERINA_HOME}" ) && -d "/usr/lib64/ballerina" ]]; then
-    export BALLERINA_HOME="/usr/lib64/ballerina/"$(ls -1 /usr/lib64/ballerina/ | grep 'ballerina[0-9\.a-zA-Z]*' | head -1)
+    export BALLERINA_HOME="/usr/lib64/ballerina/"$(ls -1t /usr/lib64/ballerina/ | grep 'ballerina[0-9\.a-zA-Z]*' | head -1)
 fi
 
 JAVA_HOME=$BALLERINA_HOME/bre/lib/jre1.8.0_172
--- a/resources/executables/linux/ballerina
+++ b/resources/executables/linux/ballerina
@@ -30,6 +30,11 @@
 # -----------------------------------------------------------------------------
 
 [ -f /etc/profile.d/wso2.sh ] && source /etc/profile.d/wso2.sh
+
+if [[ ( -z "${BALLERINA_HOME}" || ! -d "${BALLERINA_HOME}" ) && -d "/usr/lib64/ballerina" ]]; then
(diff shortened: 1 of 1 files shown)


resources/executables/linux/ballerina
```diff
@@ -30,6 +30,11 @@
 # -----------------------------------------------------------------------------
 
 [ -f /etc/profile.d/wso2.sh ] && source /etc/profile.d/wso2.sh
+
+if [[ ( -z "${BALLERINA_HOME}" || ! -d "${BALLERINA_HOME}" ) && -d "/usr/lib64/ballerina" ]]; then
+    export BALLERINA_HOME="/usr/lib64/ballerina/"$(ls -1 /usr/lib64/ballerina/ | grep 'ballerina[0-9\.a-zA-Z]*' | head -1)
```
keizer619: Currently installers keep add new files for each release upon installation of each package. Would this `$(ls -1 /usr/lib64/ballerina/ | grep 'ballerina[0-9\.a-zA-Z]*' | head -1)` conflicts when there are several files for future releases? (eg:- ballerina-0.990.0, ballerina-0.991.0,  ballerina-0.991.0)

Also keeping distribution in lib64 directory on RHEL has been done via rpm and for ubuntu we keep distribution in lib directory via deb. 

As these locations are specific for OS and package manager, shall we leave these changes in installers and not include in this ballerina shell file as its a generic? WDYT?

abhishekkr: The path capture wouldn't be a problem for multiple distribution, but now that you mentioned I have updated it to pick latest distribution placed there, earlier it was lexical first pick.

I'm not sure where package manager repo code is, else I would have updated the resolution of BALLERINA_HOME there itself.

PR #224, opened 2019-12-06
Include progress bar

keizer619:

### DIFF
--- a/ballerina-command/src/main/java/org/ballerinalang/command/util/ToolUtil.java
+++ b/ballerina-command/src/main/java/org/ballerinalang/command/util/ToolUtil.java
@@ -16,6 +16,8 @@
 
 package org.ballerinalang.command.util;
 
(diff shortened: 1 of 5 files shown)


ballerina-command/src/main/java/org/ballerinalang/command/util/ToolUtil.java
```diff
@@ -291,18 +292,23 @@ private static void downloadAndSetupDist(PrintStream printStream, HttpURLConnect
         }
     }
 
-    private static void downloadFile(PrintStream printStream, HttpURLConnection conn, String zipFileLocation)
-            throws IOException {
+    private static void downloadFile(PrintStream printStream, HttpURLConnection conn, String zipFileLocation,
+                                     String distribution) throws IOException {
```
warunalakshitha: Shall we make distribution varible name to common name. etc. downloadFileName

PR #227, opened 2019-12-06
Use get latest from service for update

keizer619:


ballerina-command/src/main/java/org/ballerinalang/command/util/ToolUtil.java
```diff
@@ -172,6 +172,41 @@ public static boolean use(PrintStream printStream, String distribution) {
...
+            throws IOException, KeyManagementException, NoSuchAlgorithmException {
+        String version = currentVersion;
+        SSLContext sc = SSLContext.getInstance("SSL");
+        sc.init(null, trustAllCerts, new java.security.SecureRandom());
+        HttpsURLConnection.setDefaultSSLSocketFactory(sc.getSocketFactory());
+
+        URL url = new URL(PRODUCTION_URL
+                + "/distributions/latest?version=" + currentVersion + "&type=" + type);
+        HttpURLConnection conn = (HttpURLConnection) url.openConnection();
+        conn.setRequestMethod("GET");
```
Kishanthan: Shouldn't there be constants for these strings "GET" & "SSL" ?

keizer619: Yes if there are any constants that would be better. Couldn't find any common constants. Most of the places have used string values.
https://www.codota.com/code/java/methods/java.net.HttpURLConnection/setRequestMethod
https://www.codota.com/code/java/methods/javax.net.ssl.SSLContext/getInstance

Issue #269, opened 2019-12-09
Fetching locally available version again doesn't give proper message


anupama-pathirage:
**Description:**
See the below sequence of command execution. Following versions are available locally arleady.

 jballerina-1.1.0-alpha
 jballerina-1.0.0

When we issue a fetch command again for one of the above versions it says `Using <version>`. Instead it should say something like `This version has fetched already`

Related issue https://github.com/ballerina-platform/ballerina-distribution/issues/267
```
➜  ~ ballerina dist list
Distributions available locally:

  jballerina-1.1.0-alpha
* jballerina-1.0.0

Distributions available remotely:

  jballerina-1.0.1
  jballerina-1.0.2
  jballerina-1.0.3
  jballerina-1.0.4

➜  ~ ballerina dist fetch jballerina-1.0.0
Using jballerina-1.0.0
➜  ~
➜  ~ ballerina dist fetch jballerina-1.1.0-alpha
Using jballerina-1.1.0-alpha
```

**Suggested Labels:**

**Suggested Assignees:**

**Affected Product Version:**

**OS, DB, other environment details and versions:**    

**Steps to reproduce:**

**Related Issues:**

Issue #267, opened 2019-12-09
Ballerina dist fetch switch to the fetched version before issuing ballerina dist use cmd


anupama-pathirage:
**Description:**
Refer to the below command sequence. Just with the ballerina fetch command it has switched the version. This should happen only after issuing `ballerina dist use`. It seems now the output of `ballerina dist fetch` is equivalant to  `ballerina dist pull`

```
➜  ~ ballerina dist list
Distributions available locally:

  jballerina-1.1.0-alpha

Distributions available remotely:

  jballerina-1.0.0
  jballerina-1.0.1
  jballerina-1.0.2
  jballerina-1.0.3
  jballerina-1.0.4

➜  ~ ballerina dist fetch jballerina-1.0.0
Current user does not have write permissions to /Library/Ballerina/distributions directory

➜  ~ sudo ballerina dist fetch jballerina-1.0.0
Password:
Sorry, try again.
Password:
Downloading jballerina-1.0.0 100% [=============================================================================================================================================================================] 99/99 (0:08:09 / 0:00:00)

jballerina-1.0.0 is installed. Please execute "ballerina dist use jballerina-1.0.0" to use as the default

➜  ~ ballerina dist list
Distributions available locally:

  jballerina-1.1.0-alpha
* jballerina-1.0.0

Distributions available remotely:

  jballerina-1.0.1
  jballerina-1.0.2
  jballerina-1.0.3
  jballerina-1.0.4

➜  ~ ballerina -v
Ballerina 1.0.0
Language specification 2019R3
Command 1.1.0-alpha
```

**Suggested Labels:**

**Suggested Assignees:**

**Affected Product Version:**

**OS, DB, other environment details and versions:**    

**Steps to reproduce:**

**Related Issues:**

SupunS:
For me, the fetch command say it uses the newly fetched one.
```
$ ballerina dist fetch jballerina-1.0.4
Using jballerina-1.0.4
```

Issue #253, opened 2019-12-09
Invalid ballerina dist command does not give proper error message


anupama-pathirage:
**Description:**
Try some invalid command like
`➜  ~ ballerina dist foo
`

The output is 
```
Unmatched argument at index 1: 'foo'
```

Instead, it should give some meaningful error like
```
ballerina: unknown command 'ballerina dist foo'
```

**Suggested Labels:**

**Suggested Assignees:**

**Affected Product Version:**

**OS, DB, other environment details and versions:**    

**Steps to reproduce:**

**Related Issues:**

keizer619:
Fixed
https://github.com/ballerina-platform/ballerina-distribution/pull/292/commits/cf22a64e722089a01eb0ee6b543a01eb54249d5f

Issue #252, opened 2019-12-09
`ballerina dist update` logs the same message twice


SupunS:
**Description:**
```
$ ballerina dist update              
Using jballerina-1.0.4
Using jballerina-1.0.4
```

Also this is there with `ballerina dist pull` command as well.

```
$ ballerina dist pull jballerina-1.0.2
Using jballerina-1.0.2
Using jballerina-1.0.2
```

PR #302, opened 2019-12-12
Fix Few issues related to ballerina commands

warunalakshitha:
## Purpose
Fix https://github.com/ballerina-platform/ballerina-distribution/issues/252
Fix https://github.com/ballerina-platform/ballerina-distribution/issues/248
Fix https://github.com/ballerina-platform/ballerina-distribution/issues/253
Fix https://github.com/ballerina-platform/ballerina-distribution/issues/267
Fix https://github.com/ballerina-platform/ballerina-distribution/issues/269


ballerina-command/src/main/java/org/ballerinalang/command/util/ToolUtil.java
```diff
@@ -565,9 +558,9 @@ public static void handleInstallDirPermission() {
         try {
             String installationPath = OSUtils.getInstallationPath();
             if (!new File(installationPath).canWrite()) {
-                throw ErrorUtil.createCommandException("current user does not have write permissions.\n\n" +
+                throw ErrorUtil.createCommandException("current user does not have permissions to run the command\n\n" +
                                                                "Current user does not have write permissions to " +
-                                                               "ballerina installation path : " + installationPath +
+                                                               "ballerina installation path " + installationPath +
```
keizer619: Ballerina reference has to be "Ballerina"

ballerina-command/src/main/java/org/ballerinalang/command/cmd/PullCommand.java
```diff
@@ -49,15 +49,30 @@ public void execute() {
...
-        } else if (pullCommands.size() > 1) {
+        }
+
+        if (pullCommands.size() > 1) {
             throw ErrorUtil.createUsageExceptionWithHelp("too many arguments given");
         }
+        ToolUtil.handleInstallDirPermission();
+        PrintStream printStream = getPrintStream();
+        String distribution = pullCommands.get(0);
+        String distributionType = distribution.split("-")[0];
```
Kishanthan: What if the dist name has completely wrong name? Have we handled that? Looks like this line will break in that case.

ballerina-command/src/main/java/org/ballerinalang/command/cmd/FetchCommand.java
```diff
@@ -49,13 +49,32 @@ public void execute() {
             return;
         }
 
-        if (fetchCommands == null) {
+        if (fetchCommands == null || fetchCommands.size() == 0) {
```
Kishanthan: We thought not to include fetch command for this release. Shall we remove it? 

We can keep this class, but remove where this is instantiated and mentioned in help text, etc.

ballerina-command/src/main/java/org/ballerinalang/command/cmd/PullCommand.java
warunalakshitha: You can get "" still split will not fail. null values are handled in upper code.

ballerina-command/src/main/java/org/ballerinalang/command/cmd/FetchCommand.java
keizer619: Lets remove from the main method and keep these logics

warunalakshitha: Will remove those in next PR.

ballerina-command/src/main/java/org/ballerinalang/command/util/ToolUtil.java
warunalakshitha: Already fixed in my next PR.

PR #343, opened 2019-12-16
Improve user experience and messages of the update tool

warunalakshitha:
## Purpose
$Subject


ballerina-command/src/main/java/org/ballerinalang/command/Main.java
```diff
@@ -19,7 +19,6 @@
 import org.ballerinalang.command.cmd.BCommand;
 import org.ballerinalang.command.cmd.DefaultCommand;
 import org.ballerinalang.command.cmd.DistributionCommand;
-import org.ballerinalang.command.cmd.FetchCommand;
```
keizer619: @warunalakshitha ; When removing fetch please remove fetch content from ballerina-dist.help as well

warunalakshitha: Removed the file since we use existing file

PR #346, opened 2019-12-16
Fix sorting local files

keizer619:


ballerina-command/src/main/java/org/ballerinalang/command/cmd/ListCommand.java
```diff
@@ -101,6 +102,7 @@ private static void listDistributions(PrintStream outStream) {
             listOfFiles = folder.listFiles();
             outStream.println("Distributions available locally: \n");
             List<String> installedVersions = new ArrayList<>();
+            Arrays.sort(listOfFiles);
```
MaryamZi: Shouldn't we show the latest at the top? 

And are remote version always sorted?

keizer619: @MaryamZi  Yes sorting descending is better. Currently remote versions are also sorted ascending order, lets handle that logic in the server side

Issue #352, opened 2019-12-17
Verify output of `ballerina update`


Kishanthan:
**Description:**
```
$ sudo ballerina update
Password:
Fetching latest version from remote server...
You are already in latest command version: 0.8.1
```

There are grammer issues with both the sentences above.

**Suggested Labels:**

**Suggested Assignees:**

**Affected Product Version:**

**OS, DB, other environment details and versions:**    

**Steps to reproduce:**

**Related Issues:**

MaryamZi:
Fixed with https://github.com/ballerina-platform/ballerina-distribution/pull/355

Issue #351, opened 2019-12-17
The command output messages, CLI help texts don't reflect what the update command really does


pubudu91:
When I heard about the `ballerina dist update` command, I assumed it updates to the latest version of Ballerina available. But now only I got to know that that is not the case. It only updates to the latest patch version. However, this precise notion of what the update command does it not reflected in the CLI help text. Even in the outputs of the command messages, in some places it refers to the patch version, and in some places it simply refers to as the latest Ballerina version. IMO, this could be confusing. We need to update the docs to include precisely what the update command does. We also need to update the outputs of the commands to convey precisely what it did. When I saw the references to the latest Ballerina version in the command outputs, I assumed it referred to the latest release done by the Ballerina team.

MaryamZi:
Fixed with https://github.com/ballerina-platform/ballerina-distribution/pull/355

PR #355, opened 2019-12-17
Update help text and update command output with accurate terms

MaryamZi:
## Purpose
$title.

Fixes https://github.com/ballerina-platform/ballerina-distribution/issues/351
Fixes https://github.com/ballerina-platform/ballerina-distribution/issues/352

### DIFF
--- a/ballerina-command/src/main/resources/cli-help/ballerina-pull.help
+++ b/ballerina-command/src/main/resources/cli-help/ballerina-pull.help
@@ -12,5 +12,5 @@ DESCRIPTION
 
 
 EXAMPLES
-       Fetch and use Ballerina 1.0.0.
+       Fetch and set Ballerina 1.0.0 as the active version.
           → ballerina dist pull jballerina-1.0.0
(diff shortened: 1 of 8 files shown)


ballerina-command/src/main/resources/cli-help/ballerina-pull.help
```diff
@@ -1,14 +1,16 @@
...
 
 DESCRIPTION
-       Dist-pull pulls and uses a given Ballerina version from the remote repository.
+       Dist-pull fetches a given Ballerina version from the remote repository
+       and sets it as the active version.
 
 
 EXAMPLES
-       Pull and use Ballerina 1.0.0.
+       Fetch and use Ballerina 1.0.0.
```
keizer619: @MaryamZi We have another command called "fetch" in the spec (although we removed it). Can't we keep pull or may be any other term like download?

MaryamZi: Since pull is pretty much fetch + use, and we are mentioning use too, I guess this is OK?

Issue #354, opened 2019-12-17
Should we log HTTP error codes in the output ?


Kishanthan:
**Description:**

Check the following scenario.

```
➜  ~ ballerina dist use jballerina-1.1.0-beta
Using distribution version: jballerina-1.1.0-beta

➜  ~ sudo ballerina dist update              
Fetching latest distribution version from remote server...
ballerina: server request failed. HTTP error code 404
```

The error is logged with 404 code. It is not useful for the user IMO. There are other places also in the codebase which logs HTTP error codes.

**Suggested Labels:**

**Suggested Assignees:**

**Affected Product Version:**

**OS, DB, other environment details and versions:**    

**Steps to reproduce:**

**Related Issues:**

MaryamZi:
Fixed with https://github.com/ballerina-platform/ballerina-distribution/pull/361.

PR #361, opened 2019-12-17
Use response message instead of code in error messages

MaryamZi:
## Purpose
$title.

Fixes https://github.com/ballerina-platform/ballerina-distribution/issues/354

### DIFF
--- a/ballerina-command/src/main/java/org/ballerinalang/command/util/ToolUtil.java
+++ b/ballerina-command/src/main/java/org/ballerinalang/command/util/ToolUtil.java
@@ -168,7 +168,7 @@ public class ToolUtil {
         conn.setRequestProperty("Accept", "application/json");
         if (conn.getResponseCode() != 200) {
             conn.disconnect();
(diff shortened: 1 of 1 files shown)


ballerina-command/src/main/java/org/ballerinalang/command/util/ToolUtil.java
```diff
@@ -168,8 +168,7 @@ public static boolean checkDistributionAvailable(String distribution) {
         conn.setRequestProperty("Accept", "application/json");
         if (conn.getResponseCode() != 200) {
             conn.disconnect();
-            throw ErrorUtil.createCommandException("server request failed. HTTP error code " +
-                                                           conn.getResponseCode());
+            throw ErrorUtil.createCommandException("server request failed: " + conn.getResponseMessage());
```
Kishanthan: Can the getResponseMessage be null in any case?

MaryamZi: Updated with d28f662 to use status code just in case the message is null.

PR #431, opened 2020-02-12
Exit the system with the non-zero status code if error occurs

suganyasuven:

### DIFF
--- a/resources/tools/bin/ballerina
+++ b/resources/tools/bin/ballerina
@@ -80,6 +80,7 @@ then
 
     export BALLERINA_HOME
     $BALLERINA_HOME/bin/./ballerina "$@"
+    exit $?
 fi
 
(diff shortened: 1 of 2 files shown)


lafernando:
Shall we write a test case for this?. We can use system:exec functionality to execute a "ballerina" process and check the exit code.

hevayo:
The issue was actually in the update tool script. Which means you will not get this fix with dist update. We will do an update tool release with this fix separately.

We are working on a installer test framework we can add this particular test there @keizer619

lafernando:
Okay, great.

lafernando:
@suganyasuven This issue is still there in Ballerina 1.1.4, when doing a fresh install. Did you actually test this fix?. If so, how was it tested? (manually that is).

suganyasuven:
This issue was in the update tool. This fix will be there in the next release of update tool.

lafernando:
But the update tool is there in the full package right? E.g. Deb package we download, rather than doing a dist update. So in this case, why isn't the updated Ballerina tool not there in v1.1.4? ..

PR #439, opened 2020-02-17
Update staging url and properties

NipunaRanasinghe:
## Purpose
$subject.


keizer619:
Closing this as we have built m1 as an SNAPSHOT

PR #441, opened 2020-02-19
Fix correct path in dist.xml for ballerina dist

hemikak:


hemikak:
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/21141

PR #472, opened 2020-03-21
Feature/gradle migration

lochana-chathura:
>In gradle, resource filters look for @@ instead of ${} during packaging. 
org.ballerinalang:ballerina-command-distribution:0.8.5 which is downloaded from https://maven.pkg.github.com/ballerina-platform/ballerina-update-tool, contains two files with variable ${ballerina.command.version} which are not filtered during packaging. 
They have to be converted from ${ballerina.command.version} to @ballerinaCommandVersion@ in order to filter them.
Above two files are as follow.
 ballerina-command-distribution-zip/bin/ballerina
 ballerina-command-distribution-zip/bin/ballerina.bat

>In the maven build, awslambda-extension-examples, docker-extension-examples, kubernetes-extension-examples are not copied to packages as their names declared wrong. It is fixed in new gradle build

>In the maven build, some files like ballerina.bat are copied to Linux and Mac packages also due to wrong declarations of "exclude"s. That also fixed with new gradle build.

>When generating a zip, entries can be added in two different ways; STORED or DEFLATED.
STORED entries are not compressed and DEFLATED entries are compressed. 
Maven assembly plugin uses STORED by default whereas gradle built-in zip uses DEFLATED by default. 
When built with default gradle zip, both sizes of the maven and gradle built zip files are same. Irrespective of they are being in two different compression methods. STORED and DEFLATED. But test is failed for DEFLATED zip files as the java library(net.lingala.zip4j) used in test java class cannot extract DEFLATED apparently. 
So changed the gradle zip compression method to STORED. Now all test cases are passing as net.lingala.zip4j can extract the zip files. But now gradle built zip file sizes are relatively larger although they are both in STORED format (Only the zip file size is large. Once they are extracted they are same size as maven builts ones). Also note that no difference in zip contents. Please refer CI/CD  team tread in which I have mention about this zip extraction issue.


ballerina-test/build.gradle
```diff
@@ -0,0 +1,39 @@
+ /*
+  ~ * Copyright (c) 2018, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
keizer619: Lets make this 2020

ballerina-test/build.gradle
```diff
@@ -0,0 +1,39 @@
...
+
+    useTestNG() {
+        suites 'src/test/resources/testng.xml'
+    }
+    testLogging {
+        showStandardStreams = true
+    }
+}
+
+build.dependsOn test
```
keizer619: Please add a new line

lochana-chathura:
First comment updated. thanks.

keizer619:
Closing this PR as we have included all of above commits in following merged PR
https://github.com/ballerina-platform/ballerina-distribution/pull/478

PR #510, opened 2020-06-02
Introduce common cache for balos, birs, and jars

BuddhiWathsala:
## Purpose
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/23648

## Approach
Currently, pack the socket balos and birs as follows:

```
jballerina-2.0.0-SNAPSHOT/cache/
├── balo
│   └── ballerina
│       └── socket
│           └── 0.6.0
│               ├── Ballerina.toml
│               └── socket-2020r1-java8-0.6.0.balo
└── bir
    └── ballerina
        └── socket
            └── 0.6.0
```

## Test environment
- Java 8
- Ballerina 2.0.0-SNAPSHOT
- ballerina/socket 0.6.0
- macOS

### DIFF
--- a/ballerina/build.gradle
+++ b/ballerina/build.gradle
@@ -890,167 +890,211 @@ artifacts {
     ballerinaWindowsDistribution file: ballerinaWindowsDistributionZip, builtBy: packageDistWindows
 }
 
-task testExamples {
+task testExamples() {
 
-     def baseBuildPath = "${project.rootDir}/ballerina/build/target/extracted-distributions/jballerina-tools-zip"
-     def distPath = "${baseBuildPath}/jballerina-tools-${ballerinaVersion}"
-     def bbeList = []
+    def jballerinaPack = "${project.rootDir}/ballerina/build/target/extracted-distributions/jballerina-tools-zip"
+    def baseBuildPath = "${project.rootDir}/ballerina/build/bbe-pack"
+    def distPath = "${baseBuildPath}/jballerina-tools-${ballerinaVersion}"
+    def bbeList = []
 
-     outputs.dir("${distPath}/examples/target")
-     outputs.dir("${distPath}/examples/.ballerina/repo/")
-     outputs.cacheIf { true }
+    PatternSet patternSet = new PatternSet();
+    patternSet.exclude("**/.ballerina/**");
+    patternSet.exclude("**/Ballerina.toml");
+    patternSet.exclude("**/Ballerina.lock");
+    patternSet.exclude("**/ballerina-internal.log")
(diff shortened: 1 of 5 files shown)


hevayo:
Can we pack the generated jar file also in the caching directory. 
ie
```
jballerina-2.0.0-SNAPSHOT/cache/
└── jar
    └── ballerina
        └── socket
            └── 0.6.0
                └── socket.jar
```

BuddhiWathsala:
Enabled the jar cache as requested. The build is failing because the new socket module cannot be found in GitHub packages. Once this https://github.com/ballerina-platform/module-ballerina-socket/pull/15 PR got merged socket artifacts will be available and build will pass.

ballerina/build.gradle
```diff
@@ -297,9 +315,16 @@ task packageDistZip(type: Zip) {
     }
 
     /* Standard Libraries */
+    def birCache = new File("${parentDir}/distributions/jballerina-${ballerinaVersion}/cache/bir/")
+    def baloCache = new File("${parentDir}/distributions/jballerina-${ballerinaVersion}/cache/balo/")
+    def jarCache = new File("${parentDir}/distributions/jballerina-${ballerinaVersion}/cache/jar/")
```
ThisaruGuruge, line 320: Isn't it better to have this path (up to `cache`) as a parameter, since we use it multiple places?

BuddhiWathsala: Fixed in here https://github.com/ballerina-platform/ballerina-distribution/pull/510/commits/4cac0641cf76edb31555ea7afaa9b6e35f536b8e

ballerina/build.gradle
```diff
@@ -311,11 +317,14 @@ task packageDistZip(type: Zip) {
     /* Standard Libraries */
     def birCache = new File("${parentDir}/distributions/jballerina-${ballerinaVersion}/cache/bir/")
     def baloCache = new File("${parentDir}/distributions/jballerina-${ballerinaVersion}/cache/balo/")
+    def jarCache = new File("${parentDir}/distributions/jballerina-${ballerinaVersion}/cache/jar/")
```
warunalakshitha: So are we copying stdlib jar to cache/jar directory and bre/lib both? Then there will be duplicate stdlib thin jars in the distribution.

BuddhiWathsala: Yes, currently we have to duplicate because we can't break the current implementation. After we totally moved to the new cache, we can remove that duplication.

hevayo: Till we do the packerina side changes we need to pack at both places

PR #515, opened 2020-06-10
Updated the examples for Client streaming and Bi directional streaming

daksithj:
## Purpose
$Subject
Depends on ballerina-platform/ballerina-lang#23821 being merged to master

### DIFF
--- a/examples/grpc-bidirectional-streaming/grpc_bidirectional_streaming_service.bal
+++ b/examples/grpc-bidirectional-streaming/grpc_bidirectional_streaming_service.bal
@@ -4,59 +4,50 @@ import ballerina/log;
 
 map<grpc:Caller> consMap = {};
 
-@grpc:ServiceConfig {
-    name: "Chat",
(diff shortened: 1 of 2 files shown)


BuddhiWathsala:
@daksithj these BBEs are [ignored during the example build](https://github.com/ballerina-platform/ballerina-distribution/blob/cb1c8401bce7721bb2f0f4ef0759ca16bbb4a624/ballerina/build.gradle#L943), shall we enable it and check whether if those are working properly.

daneshk:
> @daksithj these BBEs are [ignored during the example build](https://github.com/ballerina-platform/ballerina-distribution/blob/cb1c8401bce7721bb2f0f4ef0759ca16bbb4a624/ballerina/build.gradle#L943), shall we enable it and check whether if those are working properly?

We don't have generated stub files in the examples. We actually removed it to due avoid appearing in the bio. but we can add it in a different name so it will not pick for example page

PR #520, opened 2020-06-15
Fix socket BBE error messages

BuddhiWathsala:
## Purpose
$subject and upgrade the socket version

## Test environment
- macOS
- IDEA
- Java 8
- Ballerina 2.0.0-SNAPSHOT


BuddhiWathsala:
The build is failing because of the other BBE failures.

PR #522, opened 2020-06-15
Update task examples with metadata files

ThisaruGuruge:
## Purpose
- Added `.metatags` files
- Added links to API docs


praneesha:
Need to add the link to the API Docs module in the .description file of the BBE.

ThisaruGuruge:
> 

They are already added

PR #529, opened 2020-06-16
Update csv-io example with metadata files

MadhukaHarith92:

### DIFF
--- a/examples/csv-io/csv_io.bal
+++ b/examples/csv-io/csv_io.bal
@@ -56,8 +56,7 @@ public function main() returns @tainted error? {
     io:println("Reading  " + srcFileName + " as a table");
     var tblResult = rCsvChannel2.getTable(Employee);
     if (tblResult is table<record {}>) {
-        table<Employee> empTable = <table<Employee>> tblResult;
-        foreach var rec in empTable {
+        foreach var rec in tblResult {
             io:println(rec);
         }
     } else {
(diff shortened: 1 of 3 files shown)


examples/csv-io/csv_io.bal
```diff
@@ -81,17 +80,13 @@ public function main() returns @tainted error? {
...
     Employee[] employees = [];
     employees[0] = {id: "1", name: "Allen", salary: 300000.0};
     employees[1] = {id: "2", name: "Wallace", salary: 200000.0};
     employees[2] = {id: "3", name: "Sheldon", salary: 1000000.0};
     foreach var employee in employees {
         var result = employeeTable.add(employee);
-        if (result is error) {
-            log:printError("Error occurred while adding data to table: ",
-                            err = result);
-        }
```
daneshk, line 94: Don't we get any error while adding to the table

MadhukaHarith92: AFAIK, with the new table syntax, the `add` method doesn't return an error and instead, it panics. See https://github.com/ballerina-platform/ballerina-lang/blob/master/langlib/lang.table/src/main/ballerina/src/lang.table/table.bal#L83

daneshk: ok, then can we mention actual return type instead of var. is it () here?

daneshk: it does not return anything. we can just remove assignment

MadhukaHarith92: Ack. Updated in a63e083

examples/csv-io/csv_io.metatags
```diff
@@ -0,0 +1,2 @@
+description: BBE on how to read/write from/to a CSV file using the CSV channel.
```
Samallama, line 1:
```suggestion
description: BBE on how to read/write from/to a CSV file using the CSV channel in Ballerina.
```

PR #530, opened 2020-06-16
Create .metatags files for BBEs

Samallama:


examples/anonymous-functions/anonymous_functions.metatags
```diff
@@ -0,0 +1,2 @@
+description: Anonymous functions in Ballerina are a syntactic shortcut for defining inline functions. It only differs from a normal function because it does not have a name.
+keywords: ballerina, ballerina by example, bbe, anonymous functions
```
pubudu91, line 2: Should we consider adding "lambda functions" to this keywords list? @MaryamZi @hasithaa

rdhananjaya: +1 for lambda, I don't think I would ever search `anonymous function` in a google seach.

MaryamZi: +1 to add "lambda functions" also.

examples/functions-with-rest-parameter/functions_with_rest_parameters.metatags
```diff
@@ -0,0 +1,2 @@
+description: ballerina, ballerina by example, bbe, rest parameter, function
+keywords: A Ballerina function can have only one REST parameter and the rest parameter can take zero or more values.
```
MaryamZi, line 2: "description" and "keywords" need to be swapped?

examples/config-api/config_api.metatags
```diff
@@ -0,0 +1,2 @@
+description: The ballerina/config module in Ballerina provides an API to read configurations from various sources such as files, environment variables and CLI parameters.
+keywords: ballerina, ballerina by example, bbe, config api
```
MaryamZi, line 2: Should we maybe use "configuration" or "config management" instead of "config api"? @pubudu91 wdyt?

examples/anonymous-functions/anonymous_functions.metatags
hasithaa: Additionally, can we rename this BBE to "lambda functions" @lafernando @lakwarus WDYT?

examples/json-record-map-conversion/json_record_map_conversion.metatags
```diff
@@ -0,0 +1,2 @@
+description: This BBE shows how to convert Ballerina records, maps, and JSON objects from one type to another.
+keywords: ballerina, ballerina by example, json, record, map
```
MaryamZi, line 2: Should we use "json object" instead
```suggestion
keywords: ballerina, ballerina by example, json object, record, map
```

examples/functions-with-rest-parameter/functions_with_rest_parameters.metatags
rdhananjaya, line 2: description and keywords list are mixed up here

examples/anonymous-functions/anonymous_functions.metatags
hasithaa: Btw Spec use word "Anonymous function", instead of "Lambda"

KRVPerera: +1 for keeping both

pubudu91: I think we better keep the name of the BBE as "Anonymous functions". But for purposes like this, I think we can use the term lambda function since it's a commonly used name for anonymous functions.

examples/config-api/config_api.metatags
pubudu91: Maybe add those two words as additional keywords?

examples/functions/functions.metatags
```diff
@@ -0,0 +1,2 @@
+description: Functions in Ballerina operate in the same way as in any other language. It is a mechanism to create a reusable unit of functionality within a program.
```
pubudu91, line 1: This description needs to be updated. See the updated description in the corresponding BBE.

examples/json-record-map-conversion/json_record_map_conversion.metatags
pubudu91: For this context, it's  fine to just say json right?

examples/taint-checking/taint_checking.metatags
```diff
@@ -0,0 +1,2 @@
+description: This BBE demonstrates how taint checking ensures that programs written in Ballerina are inherently secure.
+keywords: ballerina, ballerina by example, taint checking, security
```
pubudu91, line 2: Should we consider adding something like "secure programming"? @rdhananjaya

examples/closures/closures.metatags
```diff
@@ -0,0 +1,2 @@
+description: A closure in Ballerina is an inner anonymous function that has visibility to the scope of its enclosing environment.
```
pubudu91, line 1: Need to rethink this. Not exactly the correct description of a closure IMO

examples/anonymous-functions/anonymous_functions.metatags
hasithaa, line 2:
```suggestion
keywords: ballerina, ballerina by example, bbe, anonymous functions, lambda functions
```

examples/taint-checking/taint_checking.metatags
rdhananjaya: +1, in addition, if it's ok we should include taint error messages here as most probably people will initiate a search for taint analysis because they encounter a taint error.

examples/closures/closures.metatags
rdhananjaya: +1 the description + example should also be updated right, maybe add a comment to that Anjana's spread sheet.

examples/json-arrays/json_arrays.metatags
```diff
@@ -0,0 +1,2 @@
+description: JSON array literals are written in the same way as Ballerina arrays. JSON array values must be of the JSON types: string, number, object, array, boolean or null
```
KRVPerera, line 1: Do we need a full stop at the end of the sentence?

examples/config-api/config_api.metatags
Samallama, line 2:
```suggestion
keywords: ballerina, ballerina by example, bbe, config api, config management, configuration
```

examples/functions-with-rest-parameter/functions_with_rest_parameters.metatags
```diff
@@ -0,0 +1,2 @@
+description: ballerina, ballerina by example, bbe, rest parameter, function
```
Samallama, line 1:
```suggestion
description: A Ballerina function can have only one REST parameter and the rest parameter can take zero or more values.
```

examples/functions-with-rest-parameter/functions_with_rest_parameters.metatags
Samallama, line 2:
```suggestion
keywords: ballerina, ballerina by example, bbe, rest parameter, function
```

examples/json-arrays/json_arrays.metatags
Samallama, line 1:
```suggestion
description: JSON array literals are written in the same way as Ballerina arrays. JSON array values must be of the JSON types: string, number, object, array, boolean or null.
```

examples/json-record-map-conversion/json_record_map_conversion.metatags
Samallama: which should we add here?

examples/taint-checking/taint_checking.metatags
Samallama: What are the error messages?

Samallama, line 2:
```suggestion
keywords: ballerina, ballerina by example, taint checking, security, secure programming
```

examples/functions/functions.metatags
Samallama: What can we do for now?

Samallama: Shall we delete it and add when we have the proper description?

pubudu91: I think we can just take the description from https://github.com/ballerina-platform/ballerina-distribution/pull/621/files right?

Samallama, line 1:
```suggestion
description: A function in Ballerina is a mechanism to create a reusable unit of functionality within a program.
```

Samallama: Oh right, edited!

examples/json-record-map-conversion/json_record_map_conversion.metatags
pubudu91: @MaryamZi

examples/closures/closures.metatags
pubudu91: How about something like the following?
```suggestion
description: In Ballerina, a closure is created to capture the variables of the enclosing environment when a locally defined construct (e.g., a local anonymous function) refers to them.
```

examples/json-record-map-conversion/json_record_map_conversion.metatags
MaryamZi: My original suggestion was because we can only convert a JSON object to a record or a map. Would we lose anything by making this "json object"? Even just "json" is fine.

pubudu91: Not exactly sure. I was just wondering if our keyword is "json object", whether it'd be a match for just "json". IMO, no need to go into nuances in terms for this purpose

PR #534, opened 2020-06-16
Update math description with metadata files

MadhukaHarith92:


daneshk:
Shall we add API Doc link for the code comments as well?

MadhukaHarith92:
> Shall we add API Doc link for the code comments as well?

Added in 452a59f

PR #548, opened 2020-06-17
Add raw template BBE

pubudu91:

### DIFF
--- a/examples/raw-template/raw_template.bal
+++ b/examples/raw-template/raw_template.bal
@@ -1,0 +1,63 @@
+import ballerina/lang.'object;
+import ballerina/io;
+
+public function main() {
+    string name = "Ballerina";
+    // Create a raw template with the `name` variable interpolated in the
+    // literal.
+    'object:RawTemplate template = `Hello ${name}!!!`;
+    // The string literal portions of the raw template are accessible
+    // through the `strings` field. The `strings` field is a list of
+    // `string` values. The string literal portions of the raw template
+    // are separated by the interpolations. If there are `n` insertions,
+    // there will be `n+1` strings.
+    io:println(template.strings);
+    // The `insertions` field holds the values obtained by evaluating
+    // the expressions specified in the interpolations of the raw template.
+    // In this case, there is just one interpolation: a reference to the
+    // variable, `name`.
+    io:println(template.insertions[0]);
+
+    Person[] people = [
+        {name: "John Doe", age: 25},
+        {name: "Jane Doe", age: 20}
+    ];
+
+    Query[] queries = [];
+    int i = 0;
+
+    foreach var p in people {
+        queries[i] = (`INSERT INTO Details VALUES (${p.name}, ${p.age})`);
+        i += 1;
+    }
+
+    io:println(buildQuery(queries[0]));
+    io:println(buildQuery(queries[1]));
+}
+
+// A user can define their own raw template types as well. To define a
+// raw template type,<br>
+// 1) The type should be an abstract subtype of `object:RawTemplate` type<br>
+// 2) It should only have the two fields `strings` and `insertions`<br>
+// 3) It should not have method declarations
+type Query abstract object {
+    public string[] strings;
+    // The type of the fields can be used to constrain the raw template
+    // in ways such as limiting the number of interpolations/strings
+    // that can be specified, ensuring that the interpolations are of
+    // correct types etc.
+    public [string, int] insertions;
+};
+
+type Person record {|
+    string name;
+    int age;
+|};
+
+function buildQuery(Query qu) returns string {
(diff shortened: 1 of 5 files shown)


pubudu91:
Need to add the metadata tags file to this. Don't merge yet

examples/raw-template/raw_template.bal
```diff
@@ -0,0 +1,63 @@
+import ballerina/lang.'object;
+import ballerina/io;
```
MaryamZi, line 2:
```suggestion
import ballerina/io;
import ballerina/lang.'object;
```

examples/raw-template/raw_template.metatags
```diff
@@ -0,0 +1,2 @@
+description: Unlike in string templates, raw templates have the flexibility of separately accessing the string and value portions of the expression. This example demonstrates how you can make use of that added flexibility.
+keywords: ballerina, ballerina by example, bbe, templates, raw template
```
MaryamZi, line 2: @Samallama hope this is OK.

examples/raw-template/raw_template.bal
```diff
@@ -0,0 +1,63 @@
...
+        {name: "Jane Doe", age: 20}
+    ];
+
+    Query[] queries = [];
+    int i = 0;
+
+    foreach var p in people {
+        queries[i] = (`INSERT INTO Details VALUES (${p.name}, ${p.age})`);
+        i += 1;
+    }
```
MaryamZi, line 32: Shall we do 

```suggestion

    foreach var p in people {
        queries.push(`INSERT INTO Details VALUES (${p.name}, ${p.age})`);
    }
```

instead?

examples/raw-template/raw_template.bal
```diff
@@ -0,0 +1,63 @@
...
+// 2) It should only have the two fields `strings` and `insertions`<br>
+// 3) It should not have method declarations
+type Query abstract object {
+    public string[] strings;
+    // The type of the fields can be used to constrain the raw template
+    // in ways such as limiting the number of interpolations/strings
+    // that can be specified, ensuring that the interpolations are of
+    // correct types etc.
+    public [string, int] insertions;
+};
```
MaryamZi, line 50: Shouldn't this comment come before `strings`?

```suggestion
    // The types of the fields can be used to enforce constraints 
    // on the raw template.
    // For example, a fixed-length list can limit the number of 
    // strings and insertions.
    public string[3] strings;
    // The type of `insertions` can enforce the insertions to
    // be of particular types.
    public [string, int] insertions;
};
```

examples/raw-template/raw_template.out
```diff
@@ -0,0 +1,7 @@
+# To run this sample, navigate to the directory that contains the
+# `.bal` file, and execute the `ballerina run` command.
+$ ballerina run string_template.bal
```
MaryamZi, line 3:
```suggestion
$ ballerina run raw_template.bal
```

examples/raw-template/raw_template.bal
pubudu91: This is what I initially tried to do. But ran in to this: https://github.com/ballerina-platform/ballerina-lang/issues/24268 :sweat_smile:

examples/raw-template/raw_template.bal
pubudu91: But wouldn't it then show the comment against the string field? Put it for the insertions field since it's a different type from the type defined in `RawTemplate`

examples/raw-template/raw_template.bal
MaryamZi: :sweat_smile:

examples/raw-template/raw_template.bal
MaryamZi: But it's common to both right? Using a fixed length list?

What I thought was, if we have it for `strings`, when following the flow, when we get to insertions we still know it to apply it for insertions too. 

In the current approach though we are saying something applies for `strings` too after the definition of `strings`.

examples/raw-template/raw_template.bal
```diff
@@ -0,0 +1,67 @@
...
+// 2) It should only have the two fields `strings` and `insertions`<br>
+// 3) It should not have method declarations
+type Query abstract object {
+    // The value space of the raw template type can be constrained to a certain
+    // extent by selecting suitable types for the `strings` and `insertions`
+    // fields. For example,<br>
+    // - Using a fixed-length array type to limit the number of
+    // strings/interpolations in the raw template.<br>
+    // - Using a tuple type to ensure the strings/interpolations at a given
+    // position in the raw template expression is of the correct, expected
```
MaryamZi, line 50:
```suggestion
    // position in the raw template expression are of the correct, expected
```

PR #565, opened 2020-06-18
Fix JSON/Record/Map Conversion sample

suganyasuven:
## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

### DIFF
--- a/examples/json-record-map-conversion/json_record_map_conversion.bal
+++ b/examples/json-record-map-conversion/json_record_map_conversion.bal
@@ -14,6 +14,10 @@ type Movie record {
     Person writer;
 };
 
+type j1 json;
+
+type map1 map<anydata>;
+
 public function main() {
     Movie theRevenant = {
         title: "The Revenant",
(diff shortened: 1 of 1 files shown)


examples/json-record-map-conversion/json_record_map_conversion.bal
```diff
@@ -28,16 +32,16 @@ public function main() {
     // This example shows how you can convert a record to a JSON object.
     // This conversion could return an error because it may not be possible
     // to convert some data types that are defined in the record to JSON.
-    json|error j = json.constructFrom(theRevenant);
+    json|error j = theRevenant.cloneWithType(j1);
```
niveathika, line 35:
```suggestion
    json|error j = theRevenant.cloneWithType(json);
```
Wouldn't this work?

examples/json-record-map-conversion/json_record_map_conversion.bal
```diff
@@ -14,6 +14,10 @@ type Movie record {
     Person writer;
 };
 
+type j1 json;
+
+type map1 map<anydata>;
```
pubudu91, line 19: Better if we do not name types like this. Usually, a type name starts with a capital letter and don't use numbers. We use names that convey what that type is about. How about something like `MapAnydata`?

examples/json-record-map-conversion/json_record_map_conversion.bal
pubudu91: This should work. You'll only have to define a type for types like unions.

PR #566, opened 2020-06-18
Add BBE for `never` type

dulajdilshan:
$subject

### DIFF
--- a/examples/all-bbes.json
+++ b/examples/all-bbes.json
@@ -216,6 +216,10 @@
                 "url": "any-type"
             },
             {
+                "name": "Never Type",
+                "url": "never-type"
+            },
+            {
                 "name": "Readonly Type",
                 "url": "readonly-type"
             },
--- a/examples/never-type/never_type.bal
+++ b/examples/never-type/never_type.bal
@@ -1,0 +1,32 @@
+type Person record {
+    string name;
+    int age;
+};
+
+type Student record {
+    string name;
+    // An optional field can be of type `never`, but a value can never be assigned to such a field.
+    never gender?;
+};
+
+// This function never returns. So we have `never` as the return type.
+function somefunction() returns never {
+    panic error("Invalid");
+}
+
+public function main() {
+    // This is an example of using `never` as the type parameter in `xml`.
+    // `xml<never>` describes the `xml` type that has no constituents,
+    // i.e., the empty `xml` value.
+    xml<never> xmlValue = <xml<never>> 'xml:concat();
+
+    // You can define a mapping value with `never` as its constraint.
+    // But you can never add members to this map.
+    map<never> someMap = {};
+
+    // You can specify a key-less table with the `never` type as the key constraint.
+    table<Person> key<never> personTable = table [
+        {name: "John", age: 23},
+        {name: "Paul", age: 25}
+    ];
+}
--- a/examples/never-type/never_type.description
+++ b/examples/never-type/never_type.description
@@ -1,4 +1,4 @@
 // The `never` type represents the type which has no values.
 // This can be useful to describe the return type of a function, if the function
-// never returns. As no value can ever belong to `never` type, variables of type `never`
+// never returns. As no value can ever belong to the `never` type, variables of type `never`
 // cannot be declared.
--- a/examples/never-type/never_type.description
+++ b/examples/never-type/never_type.description
@@ -1,0 +1,4 @@
+// The `never` type represents the type which has no values.
+// This can be useful to describe the return type of a function, if the function
+// never returns. As no value can ever belong to `never` type, variables of type `never`
+// cannot be declared.
--- a/examples/never-type/never_type.metatags
+++ b/examples/never-type/never_type.metatags
@@ -1,0 +1,2 @@
+description: Never type can be used to represent the type which has no values. This example demonstrates how you can use never type in some scenarios.
+keywords: ballerina, ballerina by example, bbe, type, never type
--- a/examples/never-type/never_type.out
+++ b/examples/never-type/never_type.out
@@ -1,0 +1,3 @@
+# To run this sample, navigate to the directory that contains the
+# `.bal` file and issue the `ballerina run` command.
+$ ballerina run never_type.bal


examples/all-bbes.json
```diff
@@ -222,6 +222,10 @@
             {
                 "name": "Let Expression",
                 "url": "let-expression"
+            },
+            {
+                "name": "Never Type",
+                "url": "never-type"
```
pubudu91, line 228: Shall we add this after the `any` type, so that the types are grouped together?

examples/never-type/never_type.bal
```diff
@@ -0,0 +1,44 @@
+import ballerina/io;
+import ballerina/lang.'xml as xmllib;
```
pubudu91, line 2: Better if use a quoted identifier for the module prefix. We usually don't use module aliases in lang BBEs

examples/never-type/never_type.description
```diff
@@ -0,0 +1,4 @@
+// The `never` type represents the type of values that never occur.
```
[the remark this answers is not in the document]
pubudu91: On second thoughts, shall we phrase this in a different way? This doesn't convey the idea that the value space of `never` is empty. How about something like: "The `never` type represents the type which has no value."

examples/never-type/never_type.description
```diff
@@ -0,0 +1,4 @@
...
+// This can be useful to describe the return type of a function, if the function never returns.
+// You can use `never` type as a type parameter. As no value can ever belong to `never` type, you
```
pubudu91, line 3: Didn't get what's meant by the following: "You can use `never` type as a type parameter"

examples/never-type/never_type.description
```diff
@@ -0,0 +1,4 @@
...
+// can not declare or assign any value to it.
```
pubudu91, line 4: How about, "As no value can ever belong to `never` type, variables of type `never` cannot be declared".

examples/never-type/never_type.bal
```diff
@@ -0,0 +1,44 @@
...
+    never gender?;
+};
+
+// This function never returns. So we have return type of `never`.
+function validateAndPanic(int age) returns never {
+    if (age < 18) {
+        error e = error("Invalid Age");
+        panic e;
+    }
+}
```
pubudu91, line 21: Is it correct to say that this function never returns? If age >= 18, this function completes execution normally. Doesn't that mean it returns? @hasithaa

examples/never-type/never_type.bal
```diff
@@ -0,0 +1,44 @@
...
+
+public function main() {
+    // We can call this function, but can not assign to a value.
+    validateAndPanic(20);
```
pubudu91, line 25: cannot assign to a variable

examples/never-type/never_type.bal
```diff
@@ -0,0 +1,44 @@
...
+
+    // This is an example of use of `never` type as a parameter in `xml`. 
+    // `xml<never>` describes the `xml` type that has no constituents,
+    // i.e. the empty xml value.
+    xml<never> xmlValue = <xml<never>> xmllib:concat();
+    io:println(typeof xmlValue);
+
+    // You can initialize a mapping with `never` as its type parameter.
+    map<never> someMap = {};
+    io:println(typeof someMap);
```
pubudu91, line 36: Again, this println is not really necessary

pubudu91:
Can you add the metatags file as well? Refer to other examples on what to add there

examples/never-type/never_type.bal
```diff
@@ -0,0 +1,44 @@
...
+
+public function main() {
+    // We can call this function, but can not assign to a value.
+    validateAndPanic(20);
+    io:println(typeof validateAndPanic);
+
+    // This is an example of use of `never` type as a parameter in `xml`. 
+    // `xml<never>` describes the `xml` type that has no constituents,
+    // i.e. the empty xml value.
+    xml<never> xmlValue = <xml<never>> xmllib:concat();
```
rdhananjaya, line 31: we don't need to import xml module now as it's pre-imported.
You should be able to say `xml<never> xmlValue = <xml<never>> 'xml:concat();`

examples/all-bbes.json
rdhananjaya: +1

dulajdilshan: okay

examples/never-type/never_type.bal
dulajdilshan: ack

examples/never-type/never_type.bal
```diff
@@ -0,0 +1,44 @@
...
+import ballerina/lang.'xml as xmllib;
+
+type Person record {
+    string name;
+    int age;
+};
+
+type Student record {
+    string name;
+    // You can specify an optional field with `never` type, but can not assign a value.
```
MaryamZi, line 11: How about something like

```suggestion
    // An optional field can be of type `never`, but a value can never be assigned to such a field.
    // This can be used to ensure that a value of the particular record type will never have a field 
    // by that name.
```

@pubudu91 wdyt?

examples/never-type/never_type.bal
```diff
@@ -0,0 +1,44 @@
...
+    never gender?;
+};
+
+// This function never returns. So we have return type of `never`.
```
MaryamZi, line 15:
```suggestion
// This function never returns. So we have `never` as the return type.
```

examples/never-type/never_type.bal
dulajdilshan: Ack

examples/never-type/never_type.bal
MaryamZi: I also don't think this is correct. Not returning and not panicking is the same as returning nil.

I think we should just have a function that panics either way.

And maybe **not** call it from the main function?

examples/never-type/never_type.bal
```diff
@@ -0,0 +1,44 @@
...
+        panic e;
+    }
+}
+
+public function main() {
+    // We can call this function, but can not assign to a value.
+    validateAndPanic(20);
+    io:println(typeof validateAndPanic);
+
+    // This is an example of use of `never` type as a parameter in `xml`.
```
MaryamZi, line 28:
```suggestion
    // This is an example of using `never` as the type parameter in `xml`. 
```

examples/never-type/never_type.bal
```diff
@@ -0,0 +1,44 @@
...
+}
+
+public function main() {
+    // We can call this function, but can not assign to a value.
+    validateAndPanic(20);
+    io:println(typeof validateAndPanic);
+
+    // This is an example of use of `never` type as a parameter in `xml`. 
+    // `xml<never>` describes the `xml` type that has no constituents,
+    // i.e. the empty xml value.
```
MaryamZi, line 30:
```suggestion
    // i.e., the empty `xml` value.
```

examples/never-type/never_type.bal
```diff
@@ -0,0 +1,44 @@
...
+    xml<never> xmlValue = <xml<never>> xmllib:concat();
+    io:println(typeof xmlValue);
+
+    // You can initialize a mapping with `never` as its type parameter.
```
MaryamZi, line 34:
```suggestion
    // You can initialize a mapping with `never` as its constraint.
    // But you can never add members to this map.
```

examples/never-type/never_type.description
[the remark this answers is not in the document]
MaryamZi: Just adding a minor suggestion to @pubudu91 's.

```suggestion
// The `never` type represents the type which has no values.
```

examples/never-type/never_type.description
MaryamZi: Did you mean the `xml` and `map` scenarios? Since we have the examples, IMO we don't need to say it here, since it's not a common term.

examples/never-type/never_type.bal
dulajdilshan: Removed as xml lib is pre-imported

examples/never-type/never_type.description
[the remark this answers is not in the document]
dulajdilshan: okay

examples/never-type/never_type.description
dulajdilshan: Shall I remove that part and add it to the program file ?

dulajdilshan: Or how about 

> Also, you can use `never` type as a type parameter., such as in `table` and `xml`

?

examples/never-type/never_type.bal
dulajdilshan: I'll remove the if condition and have only the `panic error` part. That will work.

examples/never-type/never_type.description
dulajdilshan: okay

examples/never-type/never_type.bal
pubudu91: +1

examples/never-type/never_type.description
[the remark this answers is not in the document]
KRVPerera: Let's go with the suggestion for now.
But shouldn't this be `shape` rather than `value`

examples/never-type/never_type.bal
```diff
@@ -0,0 +1,44 @@
...
+    if (age < 18) {
+        error e = error("Invalid Age");
+        panic e;
+    }
+}
+
+public function main() {
+    // We can call this function, but can not assign to a value.
+    validateAndPanic(20);
+    io:println(typeof validateAndPanic);
```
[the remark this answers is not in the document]
dulajdilshan: Ack

examples/never-type/never_type.bal
dulajdilshan: ack

examples/never-type/never_type.bal
dulajdilshan: Ack

examples/never-type/never_type.bal
dulajdilshan: Ack

examples/never-type/never_type.bal
dulajdilshan: Ack. and Thanks !

examples/never-type/never_type.description
dulajdilshan: Solved

examples/never-type/never_type.bal
```diff
@@ -0,0 +1,33 @@
...
+type Student record {
+    string name;
+    // An optional field can be of type `never`, but a value can never be assigned to such a field.
+    never gender?;
+};
+
+// This function never returns. So we have `never` as the return type.
+function somefunction() returns never {
+    error e = error("Invalid Age");
+    panic e;
```
pubudu91, line 15:
```suggestion
function somefunction() returns never {
    panic error("Invalid Age");
```

examples/never-type/never_type.bal
```diff
@@ -0,0 +1,33 @@
...
+}
+
+public function main() {
+    // This is an example of using `never` as the type parameter in `xml`.
+    // `xml<never>` describes the `xml` type that has no constituents,
+    // i.e., the empty `xml` value.
+    xml<never> xmlValue = <xml<never>> 'xml:concat();
+
+    // You can initialize a mapping with `never` as its constraint.
```
MaryamZi, line 24:
```suggestion
    // You can define a mapping value with `never` as its constraint.
```

examples/never-type/never_type.bal
```diff
@@ -0,0 +1,33 @@
...
+    // But you can never add members to this map.
+    map<never> someMap = {};
+
+    // You can specify a key-less table with `never` type as the key constraint.
```
MaryamZi, line 28:
```suggestion
    // You can specify a key-less table with the `never` type as the key constraint.
```

examples/never-type/never_type.bal
```diff
@@ -0,0 +1,33 @@
...
+    table<Person> key<never> personTable = table [
+        { name: "John", age: 23 },
+        { name: "Paul", age:25 }
```
MaryamZi, line 31:
```suggestion
        {name: "John", age: 23},
        {name: "Paul", age: 25}
```

examples/never-type/never_type.description
```diff
@@ -0,0 +1,4 @@
+// The `never` type represents the type which has no value.
```
MaryamZi, line 1:
```suggestion
// The `never` type represents the type which has no values.
```

examples/never-type/never_type.description
```diff
@@ -0,0 +1,4 @@
...
+// This can be useful to describe the return type of a function, if the function
+// never returns. As no value can ever belong to `never` type, variables of type `never`
+// cannot be declared.
```
MaryamZi, line 4:
```suggestion
// never returns. As no value can ever belong to the `never` type, variables of type `never`
// cannot be declared.
```

Missing a new line too.

examples/never-type/never_type.metatags
```diff
@@ -0,0 +1,2 @@
+description: Never type can be used to represent the type which has no value. This example demonstrates how you can use never type in some scenarios.
```
MaryamZi:
```suggestion
description: Never type is the type which has no value. This example demonstrates how you can use the never type in some scenarios.
```

examples/never-type/never_type.metatags
```diff
@@ -0,0 +1,2 @@
...
+keywords: ballerina, ballerina by example, bbe, templates, never type
```
MaryamZi, line 2: templates is not relevant right?

examples/never-type/never_type.out
```diff
@@ -0,0 +1,6 @@
+# To run this sample, navigate to the directory that contains the
+# `.bal` file and issue the `ballerina run` command.
+$ ballerina run never_type.bal
+typedesc function (int) returns (never)
+typedesc xml<lang.xml:Element|lang.xml:Comment|lang.xml:ProcessingInstruction|lang.xml:Text>
+typedesc map<never>
```
MaryamZi, line 6: This is outdated right?

Also, is this going to be a BBE with no output?

examples/never-type/never_type.metatags
```diff
@@ -0,0 +1,2 @@
+description: Never type can be used to represent the type which has no values. This example demonstrates how you can use never type in some scenarios.
+keywords: ballerina, ballerina by example, bbe, templates, never type
```
pubudu91, line 2: templates is not a relevant keyword for never right?

examples/never-type/never_type.out
pubudu91, line 6: Can you update the output as well

dulajdilshan: Yeah

PR #569, opened 2020-06-18
Fix csv_io.bal for new table.add syntax

KRVPerera:
## Purpose
note $subject
`table.add` does not return an error.


KRVPerera:
Closing as already fixed

PR #576, opened 2020-06-18
Fix function related BBEs

pubudu91:

### DIFF
--- a/examples/function-pointers/function_pointers.bal
+++ b/examples/function-pointers/function_pointers.bal
@@ -5,7 +5,7 @@ public function main() {
     // The function name `test` serves as a function pointer argument in the
     // call to `foo()`. Function names can be thought of as final variables
     // since although you can use it like a regular variable, you cannot
-    // modify the value it's associated with.
+    // modify the value it is associated with.
     io:println("Answer: ", foo(10, test));
     io:println("Answer: ", foo(10, getFunctionPointer()));
 
--- a/examples/function-pointers/function_pointers.bal
+++ b/examples/function-pointers/function_pointers.bal
@@ -1,7 +1,34 @@
 import ballerina/io;
 import ballerina/lang.'int;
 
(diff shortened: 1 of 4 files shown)


examples/function-pointers/function_pointers.bal
```diff
@@ -1,41 +1,44 @@
...
 
-// The `test` function acts as a variable function pointer in the `main` function.
-function test(string s, int... x) returns float {
-    int|error y = 'int:fromString(s);
-    float f = 0.0;
+public function main() {
+    // The function name `test` serves as a function pointer argument in the
+    // call to `foo()`. Function names can be thought of as final variables
+    // since although you can use it like a regular variable, you cannot
+    // modify the value it's associated with.
```
MaryamZi, line 8:
```suggestion
    // modify the value it is associated with.
```

MaryamZi: Do we need to say this specifically though?

examples/function-pointers/function_pointers.description
```diff
@@ -1,4 +1,5 @@
-// A function pointer is a Ballerina type that allows you to use functions as variables, 
-// arguments to functions, and function return values. The name of a function serves as 
-// a pointer to that function when called from other functions or operations. The definition 
-// of the function name provides the type of the pointer in terms of a function signature.
+// Ballerina allows the user to define variables of function types. This allows you to use
```
KRVPerera, line 1: I think we start by saying `user` then the second sentence we say `you`. Maybe we can omit `you`?

MaryamZi: Or 
```suggestion
// Ballerina allows defining variables of function types. This allows you to use
```

examples/function-pointers/function_pointers.bal
pubudu91: Added for the sake of completeness, since we're drawing parallels with variables.

PR #579, opened 2020-06-18
Update directory listener BBE

daneshk:
## Purpose
$subject


examples/proto-to-ballerina/proto_to_ballerina.metatags
```diff
@@ -0,0 +1,2 @@
+description: This BBE guides how to generate Ballerina code from Protocol Buffer definition
```
Samallama, line 1:
```suggestion
description: This BBE guides how to generate Ballerina code from a Protocol Buffer definition
```

PR #590, opened 2020-06-18
Add BBEs for readonly objects and readonly record/object fields

MaryamZi:
## Purpose
$title.

### DIFF
--- a/examples/readonly-objects-and-fields/readonly_objects_and_fields.bal
+++ b/examples/readonly-objects-and-fields/readonly_objects_and_fields.bal
@@ -44,6 +44,8 @@ type DefaultController object {
 };
 
 // A non-abstract `readonly object`.
+// If the object type-descriptor includes `readonly` all the fields in the
+// object are considered to be `readonly` fields.
 type MainController readonly object {
     int id;
     string[] codes;
@@ -78,11 +80,11 @@ public function main() {
     // A non-abstract object can be `readonly` either if all of its fields are `readonly` fields, or if the object
     // is defined as a `readonly` object.
     // In either case the fields cannot be updated once the value is created.
-    // Define a value of the non-abstract object type `DefaultController` with all `readonly` fields.
+    // Define a value of the non-abstract object type `DefaultController`.
     DefaultController dc = new (1122, ["LC", "AB"]);
 
-    // Since all the fields are `readonly` fields, the object value itself is considered immutable.
-    // Thus, it can be used where a `readonly` value is expected.
+    // Since all the fields of `DefaultController` are `readonly` fields, the object value itself is considered
+    // to be immutable. Thus, it can be used where a `readonly` value is expected.
     Controller & readonly controllerOne = dc;
     io:println(controllerOne.getId());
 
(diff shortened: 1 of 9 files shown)


examples/readonly-objects-and-fields/readonly_objects_and_fields.bal
```diff
@@ -0,0 +1,96 @@
...
+// A non-abstract `readonly object`.
+type MainController readonly object {
+    int id;
+    string[] codes;
+
+    function init(int id, readonly & string[] codes) {
+        self.id = id;
+        // The expected type for `codes` is `readonly & string[]`.
+        self.codes = codes;
+    }
```
pubudu91, line 55: Should we consider explaining this a bit more? Like, how we don't need to specify each field as readonly if the object is read only. I suppose why a plain int can be assigned to `id` is covered in the readonly BBE?

examples/readonly-objects-and-fields/readonly_objects_and_fields.bal
```diff
@@ -0,0 +1,96 @@
...
+
+    Details details = emp.details;
+    // The `details` field of `emp` is immutable.
+    io:println("details is immutable: ", details.isReadOnly());
+
+    // A `readonly` intersection (`T & readonly`) can be used with an object only if it is an abstract object.
+    // A non-abstract object can be `readonly` either if all of its fields are `readonly` fields, or if the object
+    // is defined as a `readonly` object.
+    // In either case the fields cannot be updated once the value is created.
+    // Define a value of the non-abstract object type `DefaultController` with all `readonly` fields.
```
pubudu91, line 81: This last sentence sorta sounds out of place with the rest of the content, since the other content are describing the concepts while this line is all of a sudden describing the line of code. Maybe can even do without the last line

MaryamZi: How about

```suggestion
    // Define a value of the non-abstract object type `DefaultController`, of which all the fields are `readonly` fields.
```

examples/readonly-objects-and-fields/readonly_objects_and_fields.bal
pubudu91: Not repeat per se. Just to point and highlight the fields that they don't need to be marked readonly when the object itself is readonly.

examples/readonly-objects-and-fields/readonly_objects_and_fields.bal
pubudu91: No the sentence itself is ok. I meant the content before that explains the concept in a generic manner, without any reference to the code line. But then the last sentence is discussing the code, without connecting the previous content with the code line.

MaryamZi: What I was trying to do here was to emphasize that this is an object with only readonly fields. To show that we can directly use it as an immutable value.

MaryamZi: Will remove this since L84 mentions this anyway.

PR #586, opened 2020-06-18
Fix and align error related BBEs

rdhananjaya:
## Purpose
Fixes
- error handling BBE
- check BBE
- checkpanic BBE
- trap BBE
- user defined error BBE

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests
   > Details about the test cases and coverage

## Security checks
 - Followed secure coding standards in http://wso2.com/technical-reports/wso2-secure-engineering-guidelines? yes/no
 - Ran FindSecurityBugs plugin and verified report? yes/no
 - Confirmed that this PR doesn't commit any keys, passwords, tokens, usernames, or other secrets? yes/no

PR #599, opened 2020-06-18
Fix suggestions for error BBEs

MaryamZi:
## Purpose
$title.

Fixes suggestions for https://github.com/ballerina-platform/ballerina-distribution/pull/586.

### DIFF
--- a/examples/trap/trap.description
+++ b/examples/trap/trap.description
@@ -1,1 +1,1 @@
-// The trap expression stops a panic and gives access to the error value associated with the panic.
+// The `trap` expression stops a panic and gives access to the error value associated with the panic.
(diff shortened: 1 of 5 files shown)


examples/trap/trap.description
```diff
@@ -1 +1 @@
-// The trap expression stops a panic and gives access to the error value associated with the panic.
+// The `trap` expression stops a panic and gives access to the error value associated with the panic.
```
pubudu91, line 1: How about captures instead of stops?
```suggestion
// The `trap` expression captures a panic and gives access to the error value associated with the panic.
```

pubudu91: Ah spec term right?

PR #608, opened 2020-06-19
Fix table.bal format and output

KRVPerera:
## Purpose
note $subject

### DIFF
--- a/examples/table/table.bal
+++ b/examples/table/table.bal
@@ -74,7 +74,8 @@ public function main() {
 
     // `.map()` applies a function to each member of a table and returns a table of the result.
     // The resulting table will have the same keys as the argument table.
-    table<Person> personTab = employeeTab.'map(function (Employee employee) returns Person {
+    table<Person> personTab = employeeTab.'map(function (Employee employee)
+                                                            returns Person {
         return {id: employee.id, name:employee.name, age:23};
     });
     io:println("Person Table Information: ", personTab);
--- a/examples/table/table.bal
+++ b/examples/table/table.bal
@@ -29,9 +29,9 @@ public function main() {
     // The table preserves the order of its members. Iterating over the table gives the
     // members of the table sequentially.
     EmployeeTable employeeTab = table [
-      { id: 1, name: "John", salary: 300.50 },
-      { id: 2, name: "Bella", salary: 500.50 },
-      { id: 3, name: "Peter", salary: 750.0 }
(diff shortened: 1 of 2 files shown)


examples/table/table.out
```diff
@@ -10,7 +10,7 @@ Employee 1: id=1 name=John salary=300.5
 Information of the removed Employee: id=2 name=Bella salary=500.5
 Employee table keys: 1 3 4
 Employee table to list: id=1 name=John salary=300.5 id=3 name=Peter salary=750.0 id=4 name=Max salary=900.0
-Employees with salary < 400.0: John
+Employees with salary < 400.0: John
```
KRVPerera, line 13: Due to this : 
https://github.com/ballerina-platform/ballerina-distribution/pull/608/files#diff-59268ffca29146e6b7b30a20bcf9af5cL70

examples/table/table.bal
```diff
@@ -76,22 +76,22 @@ public function main() {
     // The resulting table will have the same keys as the argument table.
     table<Person> personTab = employeeTab.'map(function (Employee employee)
     returns Person {
```
pubudu91, line 78: Shall we indent this as well? Kinda looks weird.

riyafa:
Shall we not delete the description comment? They are there for a purpose. If we think it is not relevant we can say N/A. And the checklists also, we need not check them if irrelevant.

examples/table/table.bal
KRVPerera: tried to fix it : cce1c6df7e6ee30e2275f773e710bcc05a997024
Do we have a proper rule for this kind of scenarios?

pubudu91: Not exactly I think

examples/table/table.bal
```diff
@@ -74,24 +74,23 @@ public function main() {
 
     // `.map()` applies a function to each member of a table and returns a table of the result.
     // The resulting table will have the same keys as the argument table.
-    table<Person> personTab = employeeTab.'map(function (Employee employee)
-    returns Person {
-        return { id: employee.id, name:employee.name, age:23 };
-    });
+    table<Person> personTab = employeeTab.'map(function (Employee employee) returns Person {
+                                    return {id: employee.id, name:employee.name, age:23};
+                              });
```
pubudu91, line 79: Btw, I didn't mean these two lines. These were correctly formatted previously. Only the `returns` part had to be formatted

KRVPerera: Check now

PR #611, opened 2020-06-19
Add enums BBE

irshadnilam:
## Purpose
$title

### DIFF
--- a/examples/enums/enums.bal
+++ b/examples/enums/enums.bal
@@ -1,0 +1,55 @@
+import ballerina/io;
+
+// An enum is defined using a module enum declaration.
+enum Color {
+  RED,
+  GREEN,
+  BLUE
+}
+
+enum Language {
+    // An enum member can explicitly specify an associated
+    // expression.
+    ENG = "English",
+    TL = "Tamil",
+    SI = "Sinhala"
+}
+
+// An enum declaration can consist of a combination of members with or without
+// explicit associated values.
+enum Bands {
+    QUEEN,
+    PF = "Pink " + "Floyd"
+}
+
+public function main() {
+    // An enum member can be used in the same way as a string constant.
+    string skyColor = BLUE;
+    io:println("skyColor: ", skyColor);
+
+    string joined = takeStrings(QUEEN, PF);
+    io:println("joined: ", joined);
+
+    // An enum can also be used as a type.
+    Language language = getLanguage("e");
+    io:println("language: ", language);
+
+    Language sinhala = "Sinhala";
+    io:println("sinhala: ", sinhala);
+
+    ENG english = "English";
+    io:println("english: ", english);
+}
+
+function takeStrings(string str1, string str2) returns string {
+    return str1 + " and " + str2;
+}
+
+function getLanguage(string symbol) returns Language {
+    if (symbol == "e") {
+        return ENG;
+    } else if (symbol == "t") {
+        return TL;
+    }
+    return SI;
+}
--- a/examples/enums/enums.description
+++ b/examples/enums/enums.description
@@ -1,0 +1,3 @@
+// An enum member is defined as a constant. An enum member can explicitly specify a constant expression
+// as an associated value. If specified, the result of evaluating the relevant expression must be a string. If the
+// constant expression is omitted, it defaults to be the same as the identifier.
--- a/examples/enums/enums.out
+++ b/examples/enums/enums.out
@@ -1,0 +1,8 @@
+# To run this sample, navigate to the directory that contains the
+# `.bal` file, and execute the `ballerina run` command.
+$ ballerina run enums.bal
+skyColor: BLUE
+joined: QUEEN and Pink Floyd
+language: English
+sinhala: Sinhala
+english: English
--- a/examples/index.json
+++ b/examples/index.json
@@ -228,6 +228,10 @@
                 "url": "constants"
             },
             {
+                "name": "Enums",
+                "url": "enums"
+            },
+            {
                 "name": "Range Expressions",
                 "url": "range-expressions"
             },


examples/enums/enums.description
```diff
@@ -0,0 +1,4 @@
+// Enums provides a convenient syntax for declaring a union of string constants.
```
KRVPerera, line 1: shouldn't this be `An Enum`

irshadnilam: Other examples have used a similar approach:
For example, Constants : https://github.com/ballerina-platform/ballerina-distribution/blob/master/examples/constants/constants.description

@MaryamZi @pubudu91 WDYT?

KRVPerera: Is "Enums" plural or singular.. then we need to change `provides`.
Sorry for not being clear?

pubudu91: Yeah, should be provide

praneesha: +1

examples/enums/enums.bal
```diff
@@ -0,0 +1,53 @@
+import ballerina/io;
+
+// Enums are defined using `enum` modifier.
+enum Color {
+  RED,
+  GREEN,
+  BLUE
+}
+
+// The value of enum members can be overridden in the enum declaration.
```
praneesha, line 10:
```suggestion
// The value of the enum members can be overridden in the enum declaration.
```

examples/enums/enums.bal
```diff
@@ -0,0 +1,53 @@
...
+enum Language {
+    ENG = "English",
+    TL = "Tamil",
+    SI = "Sinhala"
+}
+
+// The value of enum members can be a mixture of overridden and non-overridden members.
```
praneesha, line 17:
```suggestion
// The value of the enum members can be a mixture of overridden and non-overridden members.
```

examples/enums/enums.description
praneesha: +1 for provide

examples/enums/enums.description
```diff
@@ -0,0 +1,4 @@
...
+//
+// An enum member is defined as a constant. The result of evaluating the constant expression must be a string. If the
+// constant expression is omitted, it defaults to be the same as identifier.
```
praneesha, line 4:
```suggestion
// constant expression is omitted, it defaults to be the same as the identifier.
```

examples/enums/enums.bal
```diff
@@ -0,0 +1,53 @@
+import ballerina/io;
+
+// Enums are defined using `enum` modifier.
```
pubudu91, line 3: Is it correct to call `enum` a modifier? Usually we use that term for keywords which modifies the semantics of the construct. e.g., `resource`, `remote`, `public`, `private`.

examples/enums/enums.bal
```diff
@@ -0,0 +1,53 @@
...
+}
+
+// The value of enum members can be a mixture of overridden and non-overridden members.
+enum Bands {
+    QUEEN,
+    PF = "Pink " + "Floyd"
+}
+
+public function main() {
+    // An enum member can be used in the same way as a string constant declaration.
```
pubudu91, line 24:
```suggestion
    // An enum member can be used in the same way as a string constant.
```

examples/enums/enums.bal
```diff
@@ -0,0 +1,53 @@
...
+
+function takeStrings(string str1, string str2) returns string {
+    return str1 + " and " + str2;
+}
+
+function getLanguage(string symbol) returns Language {
+    if (symbol == "e") {
+        return ENG;
+    } else if (symbol == "s") {
+        return TL;
```
pubudu91, line 50: returning the wrong enum here?

examples/enums/enums.bal
irshadnilam: Shall we change it to `The value of an enum member?`, it makes more sense imo.

irshadnilam: @pubudu91 @KRVPerera

examples/enums/enums.bal
```diff
@@ -0,0 +1,53 @@
+import ballerina/io;
+
+// Enums are defined using `enum` keyword.
```
MaryamZi, line 3: Should we say

```suggestion
// Enums are defined using a module enum declaration.
```
instead?

examples/enums/enums.bal
```diff
@@ -0,0 +1,53 @@
...
+enum Color {
+  RED,
+  GREEN,
+  BLUE
+}
+
+// The value of an enum member can be overridden in the enum declaration.
```
MaryamZi, line 10: Shall we mention the constraint that it has to be a string? Also instead of overridden I think we can say something like
```suggestion
// An enum member can explicitly specify a value.
```

examples/enums/enums.bal
```diff
@@ -0,0 +1,53 @@
...
+enum Language {
+    ENG = "English",
+    TL = "Tamil",
+    SI = "Sinhala"
+}
+
+// The value of an enum member can be a mixture of overridden and non-overridden members.
```
MaryamZi, line 17: Again, the spec doesn't speak of this as overriding right?

examples/enums/enums.bal
```diff
@@ -0,0 +1,53 @@
...
+enum Bands {
+    QUEEN,
+    PF = "Pink " + "Floyd"
+}
+
+public function main() {
+    // An enum member can be used in the same way as a string constant.
+    string colorOfSky = BLUE;
```
MaryamZi, line 25: Would `skyColor` or just `sky` be a simpler variable name?

examples/enums/enums.bal
```diff
@@ -0,0 +1,53 @@
...
+    io:println("colorOfSky: ", colorOfSky);
+
+    string joined = takeStrings(QUEEN, PF);
+    io:println("joined: ", joined);
+
+    // An enum can also be used as a type.
+    Language language =  getLanguage("e");
```
MaryamZi, line 32:
```suggestion
    Language language = getLanguage("e");
```

examples/enums/enums.description
```diff
@@ -0,0 +1,4 @@
+// Enums provide a convenient syntax for declaring a union of string constants.
+//
+// An enum member is defined as a constant. The result of evaluating the constant expression must be a string. If the
+// constant expression is omitted, it defaults to be the same as the identifier.
```
MaryamZi, line 4: What is the constant expression? :smiley: Should we say something like

```suggestion
// An enum member is defined as a constant. An enum member can explicitly specify a constant expression 
// as an associated value. If specified, the result of evaluating the relevant expression must be a string. If the
// constant expression is omitted, it defaults to be the same as the identifier.
```

examples/enums/enums.bal
MaryamZi: Yeah this didn't read right to me also.

examples/enums/enums.bal
```diff
@@ -0,0 +1,52 @@
+import ballerina/io;
+
+// Enums are defined using a module enum declaration.
```
MaryamZi, line 3:
```suggestion
// An enum is defined using a module enum declaration.
```

examples/enums/enums.bal
```diff
@@ -0,0 +1,52 @@
...
+// An enum is defined using a module enum declaration.
+enum Color {
+  RED,
+  GREEN,
+  BLUE
+}
+
+// An enum member can explicitly specify a value in the enum declaration.
+enum Language {
+    ENG = "English",
```
MaryamZi, line 12: On second thoughts, shall we move this comment to L12?

```suggestion
enum Language {
    // An enum member can explicitly specify a value in the enum declaration.
    ENG = "English",
```

irshadnilam: I think it better if it was above, as this comment also applies for the next enum declaration.

MaryamZi: But that's OK right we explain the concept here and then use it in a subsequent snippet?

Maybe we can add another comment for `Bands` like
```
// An enum declaration can consist of a combination of members with or without 
// explicit associated values.
```

@pubudu91 any thoughts?

irshadnilam: +1, changed.

examples/enums/enums.bal
```diff
@@ -0,0 +1,55 @@
...
+enum Color {
+  RED,
+  GREEN,
+  BLUE
+}
+
+enum Language {
+    // An enum member can explicitly specify an associated
+    // expression.
+    ENG = "English",
```
lafernando, line 13: Shall we use the standard tags used in representing these languages, "EN", "TA", and "SI".

examples/enums/enums.bal
```diff
@@ -0,0 +1,55 @@
...
+    QUEEN,
+    PF = "Pink " + "Floyd"
+}
+
+public function main() {
+    // An enum member can be used in the same way as a string constant.
+    string skyColor = BLUE;
+    io:println("skyColor: ", skyColor);
+
+    string joined = takeStrings(QUEEN, PF);
```
lafernando, line 30: Why is this called "takeStrings"? you are just doing a concat operation there.

examples/enums/enums.bal
```diff
@@ -0,0 +1,55 @@
...
+// explicit associated values.
+enum Bands {
+    QUEEN,
+    PF = "Pink " + "Floyd"
+}
+
+public function main() {
+    // An enum member can be used in the same way as a string constant.
+    string skyColor = BLUE;
+    io:println("skyColor: ", skyColor);
```
lafernando, line 28: Shall we put a meaningful print statement there, "Color of Sky: " or something similar, not just the variable name.

examples/enums/enums.bal
```diff
@@ -0,0 +1,55 @@
...
+// An enum is defined using a module enum declaration.
+enum Color {
+  RED,
+  GREEN,
+  BLUE
+}
+
+enum Language {
+    // An enum member can explicitly specify an associated
+    // expression.
```
lafernando, line 12: Add this part to the top line itself, since it's short enough.

MaryamZi:
I've sent a new PR #663 with @irshadnilam's changes, incorporating @lafernando's suggestions.

Shall we review and merge that instead?

PR #624, opened 2020-06-22
Update function_pointers.bal

lafernando:
Updated the code to have a simpler and a meaningful example.


examples/function-pointers/function_pointers.bal
```diff
@@ -1,44 +1,24 @@
...
+function add(int v1, int v2) returns int {
+    return v1 + v2;
+}
 
-    io:println("Answer: ", foo(10, f));
+function multiply(int v1, int v2) returns int {
+    return v1 * v2;
 }
 
 // A function pointer as a parameter. A function pointer can be invoked similar
```
praneesha, line 11:
```suggestion
// Here, the function pointer is used as a parameter. A function pointer can be invoked similar
```

examples/function-pointers/function_pointers.bal
```diff
@@ -1,44 +1,24 @@
...
-        foreach var item in x {
-            f += item * 1.0 * y;
-        }
-    } else {
-        panic y;
-    }
-    return f;
+public function main() {
+    // The function name `add` serves as a function pointer argument in the
+    // call to `process()`. Function names can be thought of as final variables
```
praneesha, line 19:
```suggestion
    // call to the `process()` function. Function names can be thought of as final variables
```

examples/function-pointers/function_pointers.bal
```diff
@@ -1,44 +1,24 @@
...
+    // since although you can use it like a regular variable, you cannot
```
praneesha, line 20:
```suggestion
    // since although you can use them like a regular variable, you cannot
```

examples/function-pointers/function_pointers.bal
```diff
@@ -1,44 +1,24 @@
...
+    // modify the value it is associated with.
```
praneesha, line 21:
```suggestion
    // modify the values they are associated with.
```

lafernando: The value we are referring to here is the function, where we say it is like a regular variable, so it's a single value we talk about there.

PR #626, opened 2020-06-23
Release swan lake preview1

keizer619:


build.gradle
```diff
@@ -28,10 +28,9 @@ ext {
     ballerinaJreArtifactsVersion = "2.0.0"
     dockerVersion = project.ballerinaVersion
     kubernetesVersion = project.ballerinaVersion
-    dataMapperVersion = project.ballerinaVersion
     awslambdaVersion = project.ballerinaVersion
     picocliVersion = "4.0.1"
-    ballerinaCommandVersion = "0.8.6-SNAPSHOT"
+    ballerinaCommandVersion = "0.8.7"
```
pubudu91, line 33: Shouldn't this have been bumped to the next development version?

keizer619: This is not released by pipeline. It was added as a dependency separately therefore we need to update manually

keizer619:
> In this PR the data mapper compiler extension related changes has been removed. We will enable them in the next release once we complete the bug fix.

This has been removed in the release tag, we are merging those changes back to master. We can enable it later :)

PR #638, opened 2020-06-29
Update type_test_expression.bal

lafernando:


examples/type-test-expression/type_test_expression.out
```diff
@@ -1,10 +1,10 @@
 # To run this sample, navigate to the directory that contains the
 # `.bal` file, and execute the `ballerina run` command below.
 $ ballerina run type_test_expression.bal
```
praneesha, line 3:
```suggestion
ballerina run type_test_expression.bal
```

PR #649, opened 2020-07-01
Update anydata_type.description

lafernando:
Update the outdated definition of 'anydata'.

### DIFF
--- a/examples/anydata-type/anydata_type.bal
+++ b/examples/anydata-type/anydata_type.bal
@@ -5,7 +5,7 @@ public function main() {
     anydata data1 = 5;
     io:println(data1);
 
-    // Before using the value of `data1` in arithmetic operations, we need to ascertain 
+    // Before using the value of `data1` in arithmetic operations, it is required to ascertain 
     // that it is indeed an `int`. A type cast or a type guard can be used for this.
     int intVal = <int> data1;
     io:println(intVal + 10);
@@ -19,4 +19,3 @@ public function main() {
     anydata dataArray = intArray;
     io:println(dataArray);
 }
-
--- a/examples/anydata-type/anydata_type.bal
+++ b/examples/anydata-type/anydata_type.bal
@@ -1,30 +1,22 @@
 import ballerina/io;
 
-// This function returns a value of the `anydata` type.
-function getValue() returns anydata {
-    string name = "cat";
-    return name;
-}
-
 public function main() {
-    // In this example, the variable named `a` of the `anydata` type holds an `int` value.
-    anydata a = 5;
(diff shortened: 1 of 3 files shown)


examples/anydata-type/anydata_type.bal
```diff
@@ -1,30 +1,22 @@
...
-    io:println(a);
+    // In this example, the variable named `data1` of the `anydata` type holds an `int` value.
+    anydata data1 = 5;
+    io:println(data1);
 
-    // Before using the value of `a` in arithmetic operations, we need to
-    // ascertain that it is indeed an `int`. A type cast or
-    // a type guard can be used for this.
-    int intVal = <int>a;
+    // Before using the value of `data1` in arithmetic operations, we need to ascertain
```
praneesha, line 8:
```suggestion
    // Before using the value of `data1` in arithmetic operations, it is required to ascertain 
```

examples/anydata-type/anydata_type.bal
```diff
@@ -1,30 +1,22 @@
...
+    // In this example, the variable named `data1` of the `anydata` type holds an `int` value.
+    anydata data1 = 5;
+    io:println(data1);
 
-    // Before using the value of `a` in arithmetic operations, we need to
-    // ascertain that it is indeed an `int`. A type cast or
-    // a type guard can be used for this.
-    int intVal = <int>a;
+    // Before using the value of `data1` in arithmetic operations, we need to ascertain 
+    // that it is indeed an `int`. A type cast or a type guard can be used for this.
```
praneesha, line 9:
```suggestion
    // that it is an actual `int`. A type cast or a type guard can be used for this.
```

examples/anydata-type/anydata_type.description
```diff
@@ -1,5 +1,4 @@
-// The `anydata` type consists of pure values whose basic type is not `error`. Thus, the `anydata` type is
-// equivalent to `()|boolean|int|float|decimal|string|(anydata|error)[]|map<anydata|error>|xml|table`.
+// The `anydata` consists of plain data (a simple value, sequence value or structured value that does not contain behavioral members at any depth).
```
praneesha, line 1:
```suggestion
// The `anydata` consists of plain data (a simple value, sequence value, or structured value that does not contain behavioral members at any depth). 
```

examples/anydata-type/anydata_type.bal
lafernando: The meaning here should be more like "that is actually an `int`".

praneesha: ok +1

PR #672, opened 2020-07-11
Update query_expression_to_stream.bal

lafernando:

### DIFF
--- a/examples/query-expression-to-stream/query_expression_to_stream.bal
+++ b/examples/query-expression-to-stream/query_expression_to_stream.bal
@@ -4,21 +4,24 @@ type Student record {
     string firstName;
     string lastName;
     int intakeYear;
-    float score;
+    float gpa;
 };
 
 type Report record {
     string name;
     string degree;
-    int expectedGradYear;
+    int graduationYear;
 };
 
 public function main() {
-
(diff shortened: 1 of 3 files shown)


examples/query-expression-to-stream/query_expression_to_stream.bal
```diff
@@ -28,18 +31,18 @@ public function main() {
     stream<Report> reportStream = stream from var student in studentList
        //The `where` clause provides a way to perform conditional execution and works similarly to an `if` condition.
```
praneesha, line 32:
```suggestion
       //The `where` clause provides a way to perform conditional execution and works similar to an `if` condition.
```

examples/query-expression-to-stream/query_expression_to_stream.description
```diff
@@ -1,4 +1,4 @@
 // Adding the `stream` keyword as a prefix to a `query expression` allows users to create streams from query expressions.
 // The first clause must be the `stream from` clause and must consist of a `select` clause as well.
-// The resulted `stream` is a stream that generates the emitted values from the select clause.
-// In this scenario, the result of the query expression is a stream with the constraint type of the select clause.
+// The resulted `stream` is a stream that generates the emitted values from the `select` clause.
```
praneesha, line 3:
```suggestion
// The resulted `stream` is a stream, which generates the emitted values from the `select` clause.
```

PR #673, opened 2020-07-11
Update query_expression_to_string.description

lafernando:

### DIFF
--- a/examples/query-expression-to-string/query_expression_to_string.bal
+++ b/examples/query-expression-to-string/query_expression_to_string.bal
@@ -4,15 +4,18 @@ type Student record {
     string firstName;
     string lastName;
     int intakeYear;
-    float score;
+    float gpa;
 };
 
 public function main() {
-
-    Student s1 = {firstName: "Alex", lastName: "George", intakeYear: 2020, score: 1.5};
-    Student s2 = {firstName: "Ranjan", lastName: "Fonseka", intakeYear: 2020, score: 0.9};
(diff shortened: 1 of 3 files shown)


examples/query-expression-to-string/query_expression_to_string.bal
```diff
@@ -22,10 +25,10 @@ public function main() {
     string students = from var student in studentList
        //The `where` clause provides a way to perform conditional execution and works similarly to an `if` condition.
```
praneesha, line 26:
```suggestion
       //The `where` clause provides a way to perform conditional execution and works similar to an `if` condition.
```

examples/query-expression-to-string/query_expression_to_string.bal
```diff
@@ -22,10 +25,10 @@ public function main() {
...
        //It can refer to variables bound by the from clause.
-       //When the `where` condition evaluates to false, the iteration skips the following clauses.
-       where student.score >= 1
-       //The concatenation of the `select` clause is evaluated for each iteration.
-       select student.firstName + " " + student.lastName + ", ";
+       //When the `where` condition evaluates to false, the current iteration is skipped.
+       where student.gpa >= 2.0
+       //The values emitted from `select` clause is concatenated to get the string result of the query statement.
```
praneesha, line 30:
```suggestion
       //The values emitted from the `select` clause are concatenated to get the string result of the query statement.
```

PR #678, opened 2020-07-14
Add stdlib integration tests

BuddhiWathsala:
## Purpose
We have to pack the stdlibs using the ballerina distribution when we move to a multi-repo architecture. At that time, we need couple of really necessary integration tests to check whether the stdlibs are compatible with each other.

https://github.com/ballerina-platform/ballerina-lang/issues/21313

## Approach
Add socket related tests as the initial step.

## Test environment
- macOS
- Java 8
- Ballerina swan lake preview 1


ballerina/build.gradle
```diff
@@ -899,6 +899,74 @@ task testExamples() {
...
+     }
+     doLast {
+         ignoreList.each { String elements ->
+             stdlibTestsList.remove("$elements")
+         }
+         stdlibTestsList.each { String stdlibTest ->
+             if (Os.isFamily(Os.FAMILY_WINDOWS)) {
+                 //TODO: Need to verify with windows
+                 exec {
+                     workingDir "${buildDir}/StdlibTestProject"
```
ThisaruGuruge, line 957: In windows `/` should be changed to `\` right? (Just asking, I'm not sure.)

stdlib-integration-tests/index.json
```diff
@@ -0,0 +1,6 @@
+[
+  {
+    "name": "Socket standard library ",
+    "path": "socket"
+  }
+]
```
ThisaruGuruge, line 6:
```suggestion
]

```

ballerina/build.gradle
BuddhiWathsala: We need to test this on Windows

PR #698, opened 2020-07-17
Update object mocking BBE with preview2 syntax

azinneera:


azinneera:
Should be opened to stage-preview branch

PR #719, opened 2020-07-25
Update xml_literal.bal

lafernando:
Clarifications in code comments.


examples/xml-literal/xml_literal.bal
```diff
@@ -27,7 +28,6 @@ public function main() {
     // The expression can be a previously-defined variable, arithmetic expressions, or even a function call.
```
praneesha, line 28:
```suggestion
    // The expression can be a previously-defined variable, an arithmetic expressions, or even a function call. 
```

PR #721, opened 2020-07-25
Update xml_functions.bal

lafernando:
Clarifications in code comments.

### DIFF
--- a/examples/xml-functions/xml_functions.bal
+++ b/examples/xml-functions/xml_functions.bal
@@ -11,30 +11,30 @@ public function main() {
 
     // Get the name of an XML element.
     xmllib:Element bookNameElem = <xmllib:Element> bookName;
-    io:println(bookNameElem.getName());
+    io:println("Element name: ", bookNameElem.getName());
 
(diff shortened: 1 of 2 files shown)


examples/xml-functions/xml_functions.bal
```diff
@@ -11,30 +11,30 @@ public function main() {
...
+    io:println("All XML elements: ", x);
 
-    // Set the children elements of an XML element.
+    // Set the child elements of an XML element.
     book.setChildren(content);
-    io:println(book);
+    io:println("Child elements set: ", book);
 
     // Strip the insignificant parts of an XML value.
     // Comment items, processing instruction items are considered insignificant.
```
praneesha, line 37:
```suggestion
    // Commented items and processing instruction items are considered as insignificant.
```

lafernando: The item mentioned here is the "comment" itself, so we can't say "commented items". I've updated the rest of the line as you suggest in the following way: "Comments and processing instruction items are considered as insignificant.". Thanks.

PR #722, opened 2020-07-27
Add BBE output verification task

ramithaYO:

### DIFF
--- a/ballerina/build.gradle
+++ b/ballerina/build.gradle
@@ -987,6 +987,319 @@ task testExamples() {
     }
 }
 
+task outputVerification() {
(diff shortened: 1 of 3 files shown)


BuddhiWathsala:
@ramithaYO @keizer619, it seems this task is a replication of `testExamples`. So, I guess, we can remove `testExamples` task with this feature.

ramithaYO:
@BuddhiWathsala IMO we need both tasks, because currently with the outputVerification task we cant check all the BBEs.

ramithaYO:
@BuddhiWathsala IMO we need both tasks, because currently with the outputVerification task we cant check all BBEs.

Issue #714, opened 2020-07-23
Add daily build to the ballerina distribution


BuddhiWathsala:
**Description:**
We can do the $subject using either
- Github action
- Jenkins job

Enabling daily build make it easy to track how recent language and stdlib changes will affect to the final distribution and BBEs.

BuddhiWathsala:
@keizer619 @ThisaruGuruge WDYT?. Shall we enable a github action as a daily build?

ThisaruGuruge:
> @keizer619 @ThisaruGuruge WDYT?. Shall we enable a github action as a daily build?

+1

PR #729, opened 2020-08-07
Enable GitHub action for daily build

BuddhiWathsala:
## Purpose
Fixes https://github.com/ballerina-platform/ballerina-distribution/issues/714

### DIFF
--- a/.github/workflows/daily-build.yml
+++ b/.github/workflows/daily-build.yml
@@ -1,0 +1,38 @@
+name: Daily build
+
+on:
+  schedule:
+    - cron:  '0 */12 * * *'
+
+jobs:
+  ubuntu-build:
+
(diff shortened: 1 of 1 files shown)


.github/workflows/daily-build.yml
```diff
@@ -0,0 +1,38 @@
+name: Daily build
+
+on:
+  schedule:
+    - cron:  '0 */12 * * *'
```
ThisaruGuruge, line 5: This means that the workflow will run every 12 hours?

BuddhiWathsala: Yes.

ThisaruGuruge: Isn't it should be every 24 hours?

PR #734, opened 2020-08-12
Add NATS dependency

BuddhiWathsala:
## Purpose
$subject

## Approach
- Add NATS dependency
- ~~Remove XSLT~~
- Upgrade stdlib dependencies


daneshk:
@BuddhiWathsala Let's add `xslt` module to the distribution for now. but not push to the Ballerina Central. Before removing it completely from the distribution, we need to communicate

aashikam:
Added org-name changes in NATS examples. `ballerinax/nats`

PR #744, opened 2020-08-14
Improve test & output verification tasks

niveathika:
## Purpose
$subject for less build time.


ballerina/build.gradle
```diff
@@ -974,354 +1030,215 @@ task testExamples() {
...
             }
         }
+        
+        boolean buildFailure = false
+        File reportDetails = new File("${baseBuildPath}/result.txt")
+        if (!failedBuildExamples.isEmpty()) {
+            println "\n\tThe following BBEs failed when building\n"
+            failedBuildExamples.each {
+                println "\t ${it}"
+                reportDetails << "Build failed: ${it} \n"
```
ramithaYO, line 1225: IMO, we don't need this because we print the failed BBEs'. WDYT?

ramithaYO (approved):

ballerina/build.gradle
ramithaYO, line 1225: IMO, we don't need this because we print the failed BBEs'. WDYT?

ramithaYO, line 1225: IMO, we don't need this because we print the failed BBEs'. WDYT?

ramithaYO, line 1225: IMO, we don't need to write the result to a file because we print the failed BBEs'. WDYT?

niveathika: +1 fixed in 78aa2bc

PR #773, opened 2020-09-09
Fix a bug in BBE verification logic

ldclakmal:
## Purpose
The existing logic is as follows:
If a BBE is in the `buildIgnoreList` it is skipped from compiling and running at all. Then the remaining BBEs are considered for further actions.
- If remaining BBEs are in the `runIgnoreList`, they are tested after compiling. (run test cases)
- If remaining BBEs are not in `runIgnoreList`, they are used for output verification after compiling.

This logic skip compiling and testing (run test cases) the BBEs which are not in the `runIgnoreList`.

The updated logic will do followings:

If a BBE is in the `buildIgnoreList` it is skipped from compiling and running at all. Then the remaining BBEs are considered for further actions.
- Compile and test (run test cases) all the remaining BBEs.
- If remaining BBEs are not in the `outputVerificationIgnoreList`, they are used for output verification also.

### DIFF
--- a/ballerina/build.gradle
+++ b/ballerina/build.gradle
@@ -844,7 +844,7 @@ task testExamples() {
     patternSet.exclude("**/ballerina-internal.log")
     inputs.files(files("${distPath}/examples").asFileTree.matching(patternSet))
 
-    // The following BBEs will not be verified by running/compiling
+    // The following BBEs will not be verified by compiling/running.
     def buildIgnoreList = [
         'proto-to-ballerina',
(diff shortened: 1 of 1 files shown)


niveathika:
@ldclakmal This will increase the build time by around 8-10 minutes. But since you pointed out this includes some more test coverage from specific BBEs.Lets merge. If build run time becomes an issue later on.. we'll reconsider.

niveathika (approved):

PR #787, opened 2020-09-14
Enable filepath and time BBEs

ThishaniLucas:


examples/filepath/tests/filepath_test.bal
```diff
@@ -17,7 +16,7 @@ public function mockPrint(any|error... s) {
 
 boolean isWindows = system:getEnv("OS") != "";
 
-@test:Config
+@test:Config {}
```
niveathika, line 19: Shall we remove this test? Since the output is verified in the build?

ThishaniLucas: done

niveathika (approved):

daneshk:
@ThishaniLucas why are we removing the test case?

ThishaniLucas:
> @ThishaniLucas why are we removing the test case?

Because both the testcase and the output verifaction are doing the same thing. Therefore, it seems we can remove either.

daneshk:
> > @ThishaniLucas why are we removing the test case?
> 
> Because both the testcase and the output verifaction are doing the same thing. Therefore, it seems we can remove either.

For output verification, we need to run it manually right?. if we have the test case, it will run and verify each build

niveathika:
@daneshk No the output verification task is added in the build itself unless there are ignored explicitly all the BBEs output will be verified against .out files.

daneshk:
> @daneshk No the output verification task is added in the build itself unless there are ignored explicitly all the BBEs output will be verified against .out files.

Got it. then it's fine

PR #795, opened 2020-09-15
Add config-api example to outputVerificationIgnoreList

aashikam:
## Purpose
$title


examples/config-api/tests/config_api_test.bal
```diff
@@ -1,44 +1,44 @@
...
-    } else {
-        test:assertFail(msg = "Failed to call the endpoint:");
-    }
-}
-
-function stopService() {
-    //test:stopServices("config-api");
-}
\ No newline at end of file
+//import ballerina/test;
```
niveathika, line 1: Any reason to comment the test out?

aashikam: needs the command ```ballerina run example.bal --b7a.config.file=/path/to/conf/file/custom-config-file-name.conf``` to run the tests. There is a discussion on this in the group.

BuddhiWathsala: @aashikam can't we do something like this https://github.com/ballerina-platform/ballerina-distribution/blob/d1d1f0d29c468a25d425be65deb25e379a9b4e09/ballerina/build.gradle#L1292

niveathika: If it cant.. lets use @test:Config{enable:false}

aashikam: The test file was recently edited to not to read it from the conf file. It should be file now.

niveathika (approved):

lafernando (approved):

BuddhiWathsala (approved):

PR #800, opened 2020-09-16
Remove log-api BBE from build ignore list

MadhukaHarith92:

### DIFF
--- a/ballerina/build.gradle
+++ b/ballerina/build.gradle
@@ -885,7 +885,6 @@ task testExamples() {
         'http-client-endpoint',
         'response-with-multiparts',
         'tracing',
-        'log-api',
         'the-main-function',
         'mutual-ssl',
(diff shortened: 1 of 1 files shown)


ballerina/build.gradle
```diff
@@ -948,6 +947,7 @@ task testExamples() {
         'rabbitmq-transaction-producer',
         'rabbitmq-transaction-consumer',
         'jdbc-complex-type-queries', // Consists date time in output
+        'log-api',
```
niveathika, line 950: Any reason this can't be verified? This BBE is printing to the console right?

MadhukaHarith92: That is because the output consists of the timestamp and that varies.

niveathika: Ack

niveathika (approved):

PR #827, opened 2020-09-21
Fix json sample

suganyasuven:
## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks


examples/json/json.bal
```diff
@@ -64,8 +64,9 @@ public function main() {
     string s = j6.toJsonString();
     json j13 = checkpanic s.fromJsonString();
     io:println(j13.toJsonString());
-
-    // The value would be equal to the original value from which the string
+    // The values of the fields of the json would be equal to the values from which the string
```
praneesha, line 67:
```suggestion
    // The values of the fields of the JSON would be equal to the values from which the string
```

praneesha (approved):

examples/json/json.bal
```diff
@@ -64,8 +64,9 @@ public function main() {
...
     json j13 = checkpanic s.fromJsonString();
     io:println(j13.toJsonString());
-
-    // The value would be equal to the original value from which the string
+    // The values of the fields of the JSON would be equal to the values from which the string
     // was created.
-    io:println(j6 == j13);
+    io:println(j6.name == j13.name);
+    io:println(j6.color == j13.color);
+    io:println(j6.price == <float>j13.price);
```
MaryamZi, line 71: I don't think the comment reflects the changes - we can't cast to `float` and check equality and say they are the same.

Can we not talk about equality and change this to something like

```ballerina
    // The result of using `.fromJsonString()` with `s` would be a JSON object (`map<json>`),
    // given that the string `s` was the result of calling `.toJsonString()` on a JSON object.
    io:println(j13 is map<json>);
```

PR #878, opened 2020-10-21
Add c2c modules to distribution

anuruddhal:
## Purpose
- Add c2c modules to distribution

### DIFF
--- a/ballerina/build.gradle
+++ b/ballerina/build.gradle
@@ -42,8 +42,8 @@ def ballerinaWindowsDistributionZip = file("$project.buildDir/distributions/ball
 task unpackBallerinaJre(type: Copy) {
     group = "unpack_dependencies"
     configurations.ballerinaJre.resolvedConfiguration.resolvedArtifacts.each { artifact ->
-            from zipTree(artifact.getFile())
-            into new File("${buildDir}/target", artifact.name+"-zip")
+        from zipTree(artifact.getFile())
+        into new File("${buildDir}/target", artifact.name + "-zip")
     }
 }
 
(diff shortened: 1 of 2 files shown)


ballerina/build.gradle
```diff
@@ -203,6 +214,20 @@ task copyOtherRepos(type: Copy) {
...
+        def artifactExtractedPath = "${buildDir}/target/extracted-distributions/" + artifact.name + "-zip"
+        into("bre/lib/") {
+            from "${artifactExtractedPath}/libs"
+        }
+        into("docs/") {
+            from "${artifactExtractedPath}/docs"
+        }
+        into("bir-cache/") {
+            from "${artifactExtractedPath}/caches/bir"
+        }
```
BuddhiWathsala, line 229: Shouldn't we need to pack these C2C library artifacts to the new packerina cache?.
```gradle
// pack to new cache
into("cache/bir/") {
    from "${artifactExtractedPath}/caches/bir"
}
into("cache/balo/") {
    from "${artifactExtractedPath}/caches/balo"
}
into("cache/jar/") {
    from "${artifactExtractedPath}/caches/jar"
}
```

@anuruddhal @pramodya1994 @hevayo

anuruddhal: No. 
The annotation processors work by copying `bir-cache` and `libs` only.  
Docker and k8s annotation processors do the same.

BuddhiWathsala: Yes, but AFAIK the new packerina implementation is going to get rid of existing caches and move on to the new cache. I just want to confirm it.

hevayo: @BuddhiWathsala You are right as per our plan we have to pack it there. But with the new package API change we need to revisit this so lets merge it.

BuddhiWathsala (approved):

PR #882, opened 2020-10-21
Fix stdlib test failures in distribution

daneshk:
## Purpose
$subject


Maninda (approved):

stdlib-integration-tests/auth/tests/integration-tests/authn_config_inheritance_test.bal
```diff
@@ -36,7 +36,7 @@ public function testNoAuthHeaders1() {
     var response = clientEP0->get("/echo1/test1", req);
     if (response is http:Response) {
         assertUnauthorized(response);
-    } else {
+    } else if (response is http:ClientError) {
```
chamil321, line 39: We don't need this type test as the casting has happened in L40. Shall we cast it rather using the `else if` in other places also?

stdlib-integration-tests/auth/tests/integration-tests/authn_config_inheritance_test.bal
```diff
@@ -60,7 +60,7 @@ public function testNoAuthHeaders3() {
     var response = clientEP0->get("/echo1/test3", req);
     if (response is http:Response) {
         assertUnauthorized(response);
-    } else {
+    } else if (response is http:ClientError) {
```
chamil321, line 63: Here too

chamil321 (changes requested):

chamil321 (approved):

PR #887, opened 2020-10-22
Fix formatting for workers

riyafa:
## Purpose
Fix formatting for workers

## Goals
Formatting will match that from the tooling

## Automation tests
N/A

## Security checks
 - Followed secure coding standards in http://wso2.com/technical-reports/wso2-secure-engineering-guidelines? no
 - Ran FindSecurityBugs plugin and verified report? no
 - Confirmed that this PR doesn't commit any keys, passwords, tokens, usernames, or other secrets? no

## Samples
N/A

## Related PRs
N/A

## Migrations (if applicable)
N/A

## Test environment
N/A
 
## Learning
N/A

### DIFF
--- a/examples/workers/workers.bal
+++ b/examples/workers/workers.bal
@@ -9,23 +9,24 @@ public function main() {
 
     // This block belongs to the `w1` worker.
     worker w1 {
-        http:Client httpClient = new ("https://api.mathjs.org");
-        var response = httpClient->get("/v4/?expr=2*3");
+        http:Client httpClient = new("https://api.mathjs.org");
+        var response = httpClient -> get("/v4/?expr=2*3");
         if response is http:Response {
             io:println("Worker 1 response: ", response.getTextPayload());
         }
     }
 
     worker w2 {
-        http:Client httpClient = new ("https://api.mathjs.org");
-        var response = httpClient->get("/v4/?expr=5*7");
+        http:Client httpClient = new("https://api.mathjs.org");
+        var response = httpClient -> get("/v4/?expr=5*7");
         if response is http:Response {
             io:println("Worker 2 response: ", response.getTextPayload());
         }
     }
 
     // Waits for both workers to finish.
-    _ = wait {w1, w2};
+    _ = wait { w1, w2 };
 
     io:println("Worker execution finished");
 }
+
--- a/examples/fork/fork.bal
+++ b/examples/fork/fork.bal
@@ -22,7 +22,7 @@ public function main() {
 
     // Workers are visible outside the `fork` as futures.
     // The `wait` action will wait for both `w1` and `w2` workers to finish.
-    record {int w1; int w2;} result = wait {w1, w2};
+    record { int w1; int w2; } result = wait { w1, w2 };
 
     // The resulting record contains returned values from each worker with
     // the field name as the worker name (if a field name is not provided).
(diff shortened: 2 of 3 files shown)


examples/workers/workers.bal
```diff
@@ -9,23 +9,24 @@ public function main() {
 
     // This block belongs to the `w1` worker.
     worker w1 {
-        http:Client httpClient = new ("https://api.mathjs.org");
-        var response = httpClient->get("/v4/?expr=2*3");
+        http:Client httpClient = new("https://api.mathjs.org");
+        var response = httpClient -> get("/v4/?expr=2*3");
```
MaryamZi, line 13: Formatting shouldn't change these AFAIK. 

@SupunS, @IrushiL, thoughts please?

examples/workers/workers.bal
```diff
@@ -9,23 +9,24 @@ public function main() {
...
+        http:Client httpClient = new("https://api.mathjs.org");
+        var response = httpClient -> get("/v4/?expr=5*7");
         if response is http:Response {
             io:println("Worker 2 response: ", response.getTextPayload());
         }
     }
 
     // Waits for both workers to finish.
-    _ = wait {w1, w2};
+    _ = wait { w1, w2 };
```
MaryamZi, line 28: This too.

examples/workers/workers.bal
riyafa: The VSCode plugin changed the formatting

examples/workers/workers.bal
riyafa: Is this an issue in the formatter?

examples/workers/workers.bal
SupunS: This new formatting follows the ballerina formatting guidelines.. So we can go ahead with those unless we find any obvious issues.

However, we don't need to add the additional newline at the end. That was due to a bug in the shared pack: https://github.com/ballerina-platform/ballerina-lang/issues/26002.

MaryamZi: Have the guidelines changed? The convention was to not use spaces around `->` for remote method calls right? Given that it is a method call, just that we use `->` instead of `.`?

For `new ()` also, given that it is like `new Foo()`, at one point we decided on having a space between `new` and `()`, I can't seem to recall any documentation though.

SupunS: There is this doc https://docs.google.com/document/d/1dQ1Z_fNHgrdGqp5Nkf2aBar97kjlo2kjIf71O34u6j0/edit as the guide. But seems the remote action is undefined.

SupunS: If the remote call is something that we have already discussed and decided, +1 to go ahead with that.
Re: `new`, yes, seems there should be a space in between, as per the doc.

MaryamZi: Noted, thanks! Should we also create issues for the formatter for these changes?

Re `->`, yeah, I believe the convention was to
- have spaces around `->` for worker send
- not have spaces around `->` for remote method calls

Looping in @anupama-pathirage too.

examples/workers/workers.bal
MaryamZi: For a mapping constructor (and I assume thereby this also) the guideline was to not have spaces after `{` and before `}`.

https://ballerina.io/learn/coding-conventions/expressions/#record-literal

examples/workers/workers.bal
riyafa: shall I create an issue for the remote actions and revert the changes? And for the new as well?

IrushiL: @MaryamZi @SupunS As per the discussion, I made the following changes in PR [1].
* `new()` changed to `new ()` in `ImplicitNewExpressionNode`
* spaces around `->` removed in `RemoteMethodCallActionNode`
* spaces after `{` and before `}` are removed in inline braces

[1] https://github.com/ballerina-platform/ballerina-lang/pull/26529

riyafa:
Closing this PR as the changes are not required

PR #888, opened 2020-10-22
Update BBEs

MaryamZi:
## Purpose
$title.

### DIFF
--- a/examples/isolated-functions/isolated_functions.description
+++ b/examples/isolated-functions/isolated_functions.description
@@ -1,4 +1,4 @@
 // An `isolated` function is a function that accesses mutable state only via its parameters.
 // Such a function can only access module level state if it is final and the type is a subtype of
 // `readonly` or `isolated object {}`. Any function or method call in an `isolated` function should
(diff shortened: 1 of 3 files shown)


examples/isolated-functions/isolated_functions.description
```diff
@@ -1,4 +1,4 @@
 // An `isolated` function is a function that accesses mutable state only via its parameters.
-// Such a function can only access module level state if it is final and immutable. Any
-// function or method call in an `isolated` function should also call an `isolated` function
-// or method.
+// Such a function can only access module level state if it is final and the type is a subtype of
+// `readonly` or `isolated object {}`. Any function or method call in an `isolated` function should
+// also call an `isolated` function or method.
```
pubudu91, line 4: The last sentence, shouldn't it be something like "Any function or method call in an `isolated` function should also be calling an `isolated` function or method."?

MaryamZi: Fixed in f80f87a, thanks!

pubudu91 (approved):

PR #929, opened 2020-11-09
Update MySQL Stored Proc call BBE

Shairam:

### DIFF
--- a/examples/mysql-call-stored-procedures/mysql_call_stored_procedures.bal
+++ b/examples/mysql-call-stored-procedures/mysql_call_stored_procedures.bal
@@ -81,6 +81,7 @@ function insertRecord(mysql:Client sqlClient) {
     if (retCall is sql:ProcedureCallResult) {
         io:println("Call stored procedure `InsertStudent` is successful : ", 
                 retCall.executionResult);
+        _ = retCall.close();
     } else {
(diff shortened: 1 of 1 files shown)


examples/mysql-call-stored-procedures/mysql_call_stored_procedures.bal
```diff
@@ -81,6 +81,11 @@ function insertRecord(mysql:Client sqlClient) {
     if (retCall is sql:ProcedureCallResult) {
         io:println("Call stored procedure `InsertStudent` is successful : ", 
                 retCall.executionResult);
+
+        error? e = retCall.close();
```
niveathika, line 85:
```suggestion
        _ = retCall.close();
```
Update in other places as well

Shairam: Yes noted. I will update them. Thank you.

niveathika:
@Shairam There is no rule that entire PR should be only one commit. Undoing the committing and force pushing would be fine for single PR. However, if it is a collaboration between multiple people, there may be conflicts. If there is anything to fix in the PR add another commit do not undo the already pushed commit

niveathika (approved):

Shairam:
@niveathika  Yes sure. I will do as suggested in the future. Thank you so much.

PR #998, opened 2020-12-07
Fix broken links in disable chunking bal

chamil321:


niveathika:
This is added in https://github.com/ballerina-platform/ballerina-distribution/pull/1002

PR #1031, opened 2020-12-14
Update task  BBEs with the latest service type changes

kalaiyarasiganeshalingam:
## Purpose
$Subject


niveathika (approved):

niveathika:
@kalaiyarasiganeshalingam Shall we update the ones in Stdlib tests also?

PR #1098, opened 2021-01-14
Add random module and remove math module

MadhukaHarith92:


examples/random/random.bal
```diff
@@ -0,0 +1,12 @@
+import ballerina/io;
+import ballerina/random;
+
+public function main() {
+    // Generate a random decimal number between 0.0 and 1.0.
+    float randomDecimal = random:createDecimal();
+    io:println("Random decimal number: ", randomDecimal);
```
ThisaruGuruge, line 7:
```suggestion

    io:println("Random decimal number: ", randomDecimal);
```

examples/random/random.bal
```diff
@@ -0,0 +1,12 @@
...
+
+    // Generate a random number between the given start(inclusive) and end(exclusive) values.
+    int|error randomInteger = random:createIntInRange(1, 100);
+    io:println("Random integer number in range: ", randomInteger);
```
ThisaruGuruge, line 11:
```suggestion

    io:println("Random integer number in range: ", randomInteger);
```

gradle.properties
```diff
@@ -10,7 +10,7 @@ stdlibEncodingVersion=1.0.7-SNAPSHOT
 stdlibIoVersion=0.5.5-SNAPSHOT
 stdlibJarraysVersion=0.9.5-SNAPSHOT
 stdlibJsonutilsVersion=1.0.5-SNAPSHOT
-stdlibMathVersion=1.0.5-SNAPSHOT
+stdlibRandomVersion=1.1.0-SNAPSHOT
```
ThisaruGuruge, line 13: Why is this `1.1.0`, not `1.0.0` ?

MadhukaHarith92: Discussed and updated the version to `0.9.0`

ThisaruGuruge (approved):

gradle.properties
ThisaruGuruge: We have to update the version in the original repo too right?

MadhukaHarith92: > We have to update the version in the original repo too right?

Yes, will update.

ThisaruGuruge (changes requested):
Let's not merge this, since we are adding this in the `fix-stdlib-tests` branch

praneesha (approved):

ThisaruGuruge (approved):

PR #1112, opened 2021-01-17
Remove stringutils from distribution

ThishaniLucas:
$subject


ThisaruGuruge (approved):
Let's merge this after the migration

PR #1125, opened 2021-01-18
Fix cloneable tostring conversion failure

mohanvive:

### DIFF
--- a/examples/immutable-values/immutable_values.bal
+++ b/examples/immutable-values/immutable_values.bal
@@ -61,7 +61,7 @@ public function main() {
     if (updateResult is error) {
         // An error should occur since `m2` is immutable.
         io:println("Error occurred on update: ",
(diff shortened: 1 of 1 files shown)


KRVPerera (approved):

examples/immutable-values/immutable_values.bal
```diff
@@ -61,7 +61,7 @@ public function main() {
     if (updateResult is error) {
         // An error should occur since `m2` is immutable.
         io:println("Error occurred on update: ",
-                   <string>updateResult.detail()["message"]);
+                   updateResult.detail()["message"].toString());
```
MaryamZi, line 64: The issue here is in the cast, because now, if the source type includes error the target type should also include error.

Shall we just do
```suggestion
                   updateResult.detail()["message"]);
```

MaryamZi (approved):

PR #1129, opened 2021-01-18
Update task BBE

kalaiyarasiganeshalingam:
## Purpose
$Subject

### DIFF
--- a/examples/task-scheduler-timer/task_scheduler_timer.bal
+++ b/examples/task-scheduler-timer/task_scheduler_timer.bal
@@ -11,7 +11,7 @@ public type CounterContext record {|
 public function main() returns error? {
     // Initializes the timer scheduler using the interval value. The delay will
     // be equal to the interval if an initial delay is not provided.
-    task:Scheduler timer = new ({
+    task:Scheduler timer = check new ({
         intervalInMillis: 1000,
         initialDelayInMillis: 0
     });
--- a/examples/task-scheduler-timer/task_scheduler_timer.bal
+++ b/examples/task-scheduler-timer/task_scheduler_timer.bal
@@ -27,7 +27,7 @@ public function main() returns error? {
     // Starts the timer.
     check timer.start();
 
-    runtime:sleep(4500);
+    runtime:sleep(4.5);
 
     // Cancels the timer. This will stop the timer and all the services
     // attached to it.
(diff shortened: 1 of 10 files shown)


examples/task-scheduler-timer/task_scheduler_timer.bal
```diff
@@ -27,7 +27,7 @@ public function main() returns error? {
     // Starts the timer.
     check timer.start();
 
-    runtime:sleep(4500);
+    runtime:sleep(4.5);
```
ThisaruGuruge, line 30: Will this fix the intermittent hanging issue of this BBE test?

examples/task-service-appointment/task_service_appointment.bal
```diff
@@ -10,7 +10,7 @@ task:AppointmentConfiguration appointmentConfiguration = {
 };
 
 // Initialize the listener using the pre-defined configurations.
-listener task:Listener appointment = new (appointmentConfiguration);
+listener task:Listener appointment = check new (appointmentConfiguration);
```
ThisaruGuruge, line 13: You don't need the `check` here, since it uses the `listener` keyword

examples/task-service-timer/task_service_timer.bal
```diff
@@ -10,7 +10,7 @@ task:TimerConfiguration timerConfiguration = {
 };
 
 // Initializes the listener using the configurations defined above.
-listener task:Listener timer = new (timerConfiguration);
+listener task:Listener timer = check new (timerConfiguration);
```
ThisaruGuruge, line 13: Sam as the above comment

examples/task-scheduler-timer/task_scheduler_timer.bal
kalaiyarasiganeshalingam: Yes. Earlier the sleep interval was in milliseconds but now, it is in seconds.

ThisaruGuruge: That wasn't the issue in the test hanging. Check this error log:

https://github.com/ballerina-platform/ballerina-distribution/runs/1719729266?check_suite_focus=true

ThisaruGuruge: Oh, I may be wrong. Let's merge this.

ThisaruGuruge (approved):

PR #1179, opened 2021-01-20
Remove jdbc and mysql from distribution

kalaiyarasiganeshalingam:
## Purpose
$Subject

### DIFF
--- a/examples/task-scheduler-appointment/task_scheduler_appointment.bal
+++ b/examples/task-scheduler-appointment/task_scheduler_appointment.bal
@@ -34,7 +34,8 @@ service object {} appointmentService = service object {
     // This resource is triggered when the appointment is due.
     remote function onTrigger() {
         reminderCount += 1;
-        io:println("Schedule is due - Reminder: ", reminderCount);
+        if (reminderCount <= 5) {
+            io:println("Schedule is due - Reminder: ", reminderCount);
+        }
     }
-
 };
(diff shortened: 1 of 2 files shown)


examples/jdbc-complex-type-queries/jdbc_complex_type_queries.bal
```diff
@@ -1,5 +1,5 @@
 import ballerina/io;
-import ballerina/jdbc;
+import ballerina/java.jdbc as jdbc;
```
anupama-pathirage, line 2: @daneshk Shouldn's this be `ballerinax/java.jdbc`?

kalaiyarasiganeshalingam: Yes. This should be `ballerinax/java.jdbc`.

examples/jdbc-query-operation/jdbc_query_operation.bal
```diff
@@ -1,5 +1,5 @@
 import ballerina/io;
-import ballerina/jdbc;
+import ballerina/java.jdbc as jdbc;
```
niveathika, line 2:
```suggestion
import ballerina/java.jdbc;
```
This we can simply leave it this, can't we? In this case also the alias will be `jdbc`

stdlib-integration-tests/sql/tests/jdbc-integration-test.bal
```diff
@@ -25,13 +25,15 @@ string dbPath = checkpanic file:getAbsolutePath("target/databases");
 string jdbcUrl = "jdbc:h2:" + dbPath + "/" + "CONNECT_DB";
 
 @test:Config {
+    enable: false
```
niveathika, line 28: Rather than removing separately, shall we add this to the ignore list in the build task?

PR #1243, opened 2021-01-22
Fix several BBEs and tests

MaryamZi:
## Purpose
$title.

### DIFF
--- a/ballerina/build.gradle
+++ b/ballerina/build.gradle
@@ -691,27 +691,17 @@ task testExamples() {
 
             // Disabling temporarily (https://github.com/ballerina-platform/ballerina-distribution/issues/1242)
             "ballerina-to-openapi",
-            "dynamic-listener",
             "xa-transactions",
             "retry-transactions",
             "non-transaction-retry",
             "local-transactions-with-handlers",
             "local-transactions",
             "log-api",
-            "async",
-            "worker-interaction",
-            "workers",
-            "fail",
-            "xml-attributes",
-            "any-type",
             "optional-type",
             "error-match-statement",
             "mapping-match-statement",
             "tuple-match-statement",
-            "foreach",
-            "while",
-            "function-pointers",
-            "modules"
+            "while"
     ]
 
     // The following BBEs will be excluded from output verification, which means verifying the output of the `.bal` file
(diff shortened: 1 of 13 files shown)


ballerina/build.gradle
```diff
@@ -691,27 +691,17 @@ task testExamples() {
...
             "local-transactions-with-handlers",
             "local-transactions",
             "log-api",
-            "async",
-            "worker-interaction",
-            "workers",
-            "fail",
-            "xml-attributes",
-            "any-type",
             "optional-type",
```
MaryamZi, line 700: Can not be enabled due to an issue in mocking. Same for the while BBE.

examples/modules/modules.bal
```diff
@@ -1,6 +1,6 @@
 // This example imports a module. You can only refer to the public symbols of
 // an imported module.
-import ballerina/math;
+import ballerina/lang.'float;
```
MaryamZi, line 3: Not the best example, since there's a predeclared import for this. Need to find a suitable alternative with a constant.

ballerina/build.gradle
```diff
@@ -691,27 +691,17 @@ task testExamples() {
...
             // Disabling temporarily (https://github.com/ballerina-platform/ballerina-distribution/issues/1242)
             "ballerina-to-openapi",
-            "dynamic-listener",
             "xa-transactions",
             "retry-transactions",
             "non-transaction-retry",
             "local-transactions-with-handlers",
             "local-transactions",
             "log-api",
-            "async",
```
warunalakshitha, line 701: Any reason why we remove these from tests?

MaryamZi: This is the ignore list. :smiley: Enabling disabled tests.

examples/any-type/tests/any_type_test.bal
```diff
@@ -10,9 +10,11 @@ string[] outputs = [];
 test:MockFunction mock_printLn = new();
 
 public function mockPrint(any|error... val) {
-    outputs.push(val.reduce(function (any|error a, any|error b) returns string => a.toString() + b.toString(), "").toString());
+    outputs.push(toString(val.reduce(function (any|error a, any|error b) returns string => toString(a) + toString(b), "")));
```
ThisaruGuruge, line 13: Line longer than 80 characters? This will fail at the `LengthTest`. Shall we reduce the length to 80 chars?

ThisaruGuruge: Oh this is a test. Ignore the comment

examples/modules/modules.bal
ThisaruGuruge: @MadhukaHarith92

PR #1246, opened 2021-01-22
Fix several BBEs and tests

MaryamZi:
## Purpose
$title.

### DIFF
--- a/examples/modules/modules.bal
+++ b/examples/modules/modules.bal
@@ -1,6 +1,6 @@
 // This example imports a module. You can only refer to the public symbols of
 // an imported module.
-import ballerina/math;
+import ballerina/lang.'float;
 
 // Declare an explicit prefix.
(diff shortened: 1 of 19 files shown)


warunalakshitha (approved):

KavinduZoysa (approved):

examples/modules/modules.bal
```diff
@@ -1,6 +1,6 @@
 // This example imports a module. You can only refer to the public symbols of
 // an imported module.
-import ballerina/math;
+import ballerina/lang.'float;
```
MaryamZi, line 3: Not the best example, since there's a predeclared import for this. Need to find a suitable alternative with a constant.

MaryamZi: Created https://github.com/ballerina-platform/ballerina-distribution/issues/1247.

PR #1255, opened 2021-01-25
Fix gRPC BBE config path and ballerina->bal change

BuddhiWathsala:
## Purpose
$subject

### DIFF
--- a/examples/grpc-secured-unary/grpc_secured_unary_service.bal
+++ b/examples/grpc-secured-unary/grpc_secured_unary_service.bal
@@ -2,15 +2,12 @@
 import ballerina/grpc;
 import ballerina/log;
 
-// The configurable path to the Ballerina home.
-configurable string balHome = "<PATH_TO_THE_BALLERINA_HOME>";
-
 // The server endpoint configuration with the SSL configurations.
 listener grpc:Listener ep = new (9090, {
     host: "localhost",
     secureSocket: {
         keyStore: {
-            path: balHome + "/bre/security/ballerinaKeystore.p12",
+            path: "../resources/ballerinaKeystore.p12",
             password: "ballerina"
         }
     }
--- a/examples/grpc-secured-unary/grpc_secured_unary_service.bal
+++ b/examples/grpc-secured-unary/grpc_secured_unary_service.bal
@@ -1,15 +1,16 @@
 // This is the server implementation of the secured connection (HTTPS) scenario.
-import ballerina/config;
 import ballerina/grpc;
 import ballerina/log;
+
+// The configurable path to the Ballerina home.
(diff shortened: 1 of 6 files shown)


HindujaB (commented):
The `.out` file still shows steps to pass config API parameters. We should change this with updated details. 
https://github.com/ballerina-platform/ballerina-distribution/blob/master/examples/grpc-secured-unary/grpc_secured_unary_service.out

examples/grpc-secured-unary/grpc_secured_unary_service.bal
```diff
@@ -1,15 +1,16 @@
 // This is the server implementation of the secured connection (HTTPS) scenario.
-import ballerina/config;
 import ballerina/grpc;
 import ballerina/log;
 
+// The configurable path to the Ballerina home.
+configurable string balHome = "<PATH_TO_THE_BALLERINA_HOME>";
```
HindujaB, line 6: Do we need a default `Config.toml` file for the test execution to pass?

BuddhiWathsala: @daneshk, what should be the better approach, 
- using the ballerina home path as a configurable variable OR
- use normal hard coded ballerina constant

BuddhiWathsala: Here I added this hardcoded `balHome` configurable variable because, for the sake of completeness of this BBE, but don't want to go into much detail like `Config.toml`. There is no point in discussing configuration here since we have a separate BBE for that and we just want to emphasize the gRPC use case here.

daneshk: Shall we do it similar to https listener example. https://github.com/ballerina-platform/ballerina-distribution/blob/master/examples/https-listener/https_listener.bal#L11

BuddhiWathsala: Fixed in here https://github.com/ballerina-platform/ballerina-distribution/pull/1255/commits/4af483997d546e843b1d8c51b23099ca04d205fc

praneesha (approved):

daneshk (approved):

BuddhiWathsala:
> The `.out` file still shows steps to pass config API parameters. We should change this with updated details.
> https://github.com/ballerina-platform/ballerina-distribution/blob/master/examples/grpc-secured-unary/grpc_secured_unary_service.out

Fixed in here https://github.com/ballerina-platform/ballerina-distribution/pull/1255/commits/4c6edf25eefbc0b730ce203415b80e919f4ff33c

HindujaB (approved):

PR #1264, opened 2021-01-26
Fix out verification failure of task scheduler BBE

kalaiyarasiganeshalingam:
## Purpose
$Subject

### DIFF
--- a/examples/task-scheduler-appointment/task_scheduler_appointment.bal
+++ b/examples/task-scheduler-appointment/task_scheduler_appointment.bal
@@ -34,7 +34,8 @@ service object {} appointmentService = service object {
     // This resource is triggered when the appointment is due.
     remote function onTrigger() {
         reminderCount += 1;
-        io:println("Schedule is due - Reminder: ", reminderCount);
+        if (reminderCount <= 5) {
+            io:println("Schedule is due - Reminder: ", reminderCount);
+        }
     }
-
 };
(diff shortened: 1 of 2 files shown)


examples/task-scheduler-appointment/task_scheduler_appointment.bal
```diff
@@ -34,7 +34,8 @@ service object {} appointmentService = service object {
     // This resource is triggered when the appointment is due.
     remote function onTrigger() {
         reminderCount += 1;
-        io:println("Schedule is due - Reminder: ", reminderCount);
+        if (reminderCount <= 5) {
```
ThisaruGuruge, line 37: What will happen if the counter goes past 5? I've seen this BBE logs reminder count like `2000+` number. Is there a way to stop this after sometime?

kalaiyarasiganeshalingam: We have recently changed the interval from milliseconds to seconds in the sleep(). Therefore, you saw that count like `2000+`  before updated with the latest changes(`runtime:sleep(9000)`. It means strand waits for 9000 seconds). Now, This has updated [to seconds](https://github.com/ballerina-platform/ballerina-distribution/blob/master/examples/task-scheduler-appointment/task_scheduler_appointment.bal#L24). So It waits for nine seconds and stops the scheduler.

ThisaruGuruge (approved):

PR #1329, opened 2021-02-16
Update testing pack to final zip pack

keizer619:

### DIFF
--- a/ballerina/build.gradle
+++ b/ballerina/build.gradle
@@ -1322,21 +1322,21 @@ packageDistZip.dependsOn packageDist
 packageDistLinux.dependsOn packageDistZip
 packageDistMac.dependsOn packageDistLinux
 packageDistWindows.dependsOn packageDistMac
-
+unzipDistForTests.dependsOn packageDistZip
 unpackStdLibs.dependsOn unpackBallerinaJre
 unpackStdLibs.dependsOn unpackJballerinaTools
 
-testExamples.dependsOn packageDist
+testExamples.dependsOn unzipDistForTests
 
 testStdlibs.dependsOn startLdapServer
-testStdlibs.dependsOn packageDist
+testStdlibs.dependsOn unzipDistForTests
 testStdlibs.finalizedBy stopLdapServer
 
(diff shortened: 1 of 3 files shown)


pubudu91 (approved):

ballerina/build.gradle
```diff
@@ -895,7 +893,7 @@ task testExamples() {
         def outputFilePath = "${directoryPath}/output.txt"
         def additionalBuildParams = ""
         if (bbe == "secured-service-with-basic-auth" || bbe == "secured-client-with-basic-auth") {
-            additionalBuildParams = "--b7a.config.file=${baseBuildPath}/${bbe}/sample-users.toml"
+            additionalBuildParams = "--b7a.config.file=${distPath}/${bbe}/sample-users.toml"
```
pubudu91, line 896: This probably need to be updated right? This is a usage of the old config API. @ldclakmal thoughts?

ballerina/build.gradle
```diff
@@ -1114,7 +1105,7 @@ task testStdlibs() {
         def exitVal
         def additionalBuildParams = ""
         if (stdlibTest == "auth" || stdlibTest == "jwt") {
-            additionalBuildParams = "--b7a.config.file=${baseBuildPath}/${stdlibTest}/tests/resources/datafiles/users.toml"
+            additionalBuildParams = "--b7a.config.file=${distPath}/${stdlibTest}/tests/resources/datafiles/users.toml"
```
pubudu91, line 1108: Same here.

ballerina/build.gradle
keizer619: Just wondering how these tests were passing, if its an invalid config API shouldn't it fail?

ldclakmal: Yes. We don't need this anymore. These are old config API usages. With the new `configurable` feature, those test cases were also removed. That's why the tests were passing.

ldclakmal, line 896:
```suggestion
```

ballerina/build.gradle
ldclakmal, line 1108:
```suggestion
```

PR #1332, opened 2021-02-17
Refactor names of the WebSocket BBEs

Bhashinee:
## Purpose
$subject.

### DIFF
--- a/examples/index.json
+++ b/examples/index.json
@@ -1002,22 +1002,23 @@
         "category": "Working over the network",
         "samples": [
             {
-                "name": "Handle Text Message with Client",
+                "name": "Handle Text Messages with Client",
                 "url": "websocket-text-client",
                 "disablePlayground": true
             },
             {
-                "name": "Handle Binary Message with Client",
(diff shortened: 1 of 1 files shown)


praneesha:
Shall we change the order as follows:

- Handle Text Messages with the Client
- Handle Text Messages with the Async Client
- Handle Binary Messages with the Client
- Handle Binary Message with the Async Client
- Listener Functionalities

@shafreenAnfar - Better if we can specify `the client` as `the default/primary/basic/sync client'. If not, shall we go with the above list?

shafreenAnfar:
I am no expert. It is up to devrel team to decide. But I feel it is better without adding `the`.

praneesha:
> I am no expert. It is up to devrel team to decide. But I feel it is better without adding `the`.

Need your feedback on the order change we did and if it is fine to specify the client as `the default/primary/basic/sync client`.

praneesha:
LGTM now.

shafreenAnfar (approved):

PR #1354, opened 2021-02-25
Add xmldata module

kalaiyarasiganeshalingam:
## Purpose
Fixing: [#407](https://github.com/ballerina-platform/ballerina-standard-library/issues/407)


gradle.properties
```diff
@@ -12,13 +12,12 @@ ballerinaCommandVersion=0.8.20
...
 stdlibIoVersion=0.6.0-alpha5-SNAPSHOT
 stdlibJavaArraysVersion=0.10.0-alpha5-SNAPSHOT
-stdlibJsonutilsVersion=1.1.0-alpha5-SNAPSHOT
 stdlibRandomVersion=0.10.0-alpha5-SNAPSHOT
 stdlibRegexVersion=0.7.0-alpha5-SNAPSHOT
 stdlibRuntimeVersion=0.6.0-alpha5-SNAPSHOT
 stdlibTaskVersion=1.2.0-alpha5-SNAPSHOT
 stdlibTimeVersion=1.1.0-alpha5-SNAPSHOT
-stdlibXmlutilsVersion=1.1.0-alpha5-SNAPSHOT
+stdlibXmldataVersion=1.1.0-alpha5-SNAPSHOT
```
ThisaruGuruge, line 20: Is this version correct? Since this is a new module, I guess the version should be `1.0.0-SNAPSHOT` or `0.1.0-SNAPSHOT` (Or something similar. The version `1.1.0-SNAPSHOT` feels like there was a `1.0.x` versions before.

kalaiyarasiganeshalingam: @daneshk Any thoughts about this?

daneshk: Let's make it as `1.0.0-SNAPSHOT`

ThisaruGuruge: Ah, but there's a problem. Since we just changed the repo name from `xmlutils` to `xmldata`, `1.0.0` version can create issues no?

ThisaruGuruge: > Ah, but there's a problem. Since we just changed the repo name from `xmlutils` to `xmldata`, `1.0.0` version can create issues no?

In that case I think `2.0.0` is a better version since this can be considered as a total revamp of the module `xmlutils` module?

kalaiyarasiganeshalingam: We have renamed the module but this is a new module, Therefore, I think we can't go with 2.0.0. Can't we go with 1.2.0?

ThisaruGuruge: Let's keep it as `1.1.0` then.

ThisaruGuruge (approved):

PR #1516, opened 2021-03-29
Include release publish workflow and improve daily builds

keizer619:
## Purpose

- Archive daily builds
- Include release publish workflow

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -1,0 +1,101 @@
+name: Publish release
+
+on:
+  workflow_dispatch:
+
+jobs:
+  publish-release:
+    name: Publish Release
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set version env variable
+        id: version-set
+        run: |
+          SHORT_VERSION=$((grep -w 'version' | cut -d= -f2) < gradle.properties
+          RELEASE_VERSION=$((grep -w 'shortVersion' | cut -d= -f2) < gradle.properties | rev | cut --complement -d- -f1 | rev)
+          echo VERSION=$RELEASE_VERSION >> $GITHUB_ENV
+          echo "::set-output name=version::$RELEASE_VERSION"
+          echo "::set-output name=sversion::SHORT_VERSION"
+      - name: Set up JDK 1.8
+        uses: actions/setup-java@v1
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-release.yml
```diff
@@ -0,0 +1,101 @@
+name: Publish release
+
+on:
+  workflow_dispatch:
+
+jobs:
+  publish-release:
+    name: Publish Release
+    runs-on: ubuntu-latest
```
pubudu91, line 9: Shall we stick to a specific version? Recently we ran in to a weird build failure which turned out to be due to `ubuntu-latest` pulling in 20.04 (it used pull 18).

keizer619: In which repo it happened? is it during release? 

Apparently most of repos are using `ubuntu-latest` including [ballerina-lang](https://github.com/ballerina-platform/ballerina-lang/blob/master/.github/workflows/publish_release.yml#L8)

.github/workflows/publish-release.yml
```diff
@@ -0,0 +1,101 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set version env variable
+        id: version-set
+        run: |
+          SHORT_VERSION=$((grep -w 'version' | cut -d= -f2) < gradle.properties
+          RELEASE_VERSION=$((grep -w 'shortVersion' | cut -d= -f2) < gradle.properties | rev | cut --complement -d- -f1 | rev)
```
pubudu91, line 16: Shouldn't these two be swapped? i.e., `RELEASE_VERSION <= 'version'` and `SHORT_VERSION <= 'shortVersion'`?

.github/workflows/publish-release.yml
Kishanthan: +1

Kishanthan: Happened for website repo, with some stuffs related to python and aws cli tools

keizer619: In that case shall we leave `ubuntu-latest` for these and keep exact version which ever failing ones.
Usually python versions in Ubuntu 18 and 20 are different and our python script may not support the latest. Ideally we should update that script to support latest

pubudu91 (approved):
Looks ok to me.

PR #1594, opened 2021-04-19
Modify example for commandline parsing

riyafa:


examples/the-main-function/the_main_function.bal
```diff
@@ -1,26 +1,49 @@
...
+|};
 
+// The `main` function accepts Person information and prints the message 
+// `repeats` number of times while prepending the `messages`.
+// The first parameter `repeats` is a required parameter and there can be 
+// 0 or more `messages`. The parameters `repeats` and `messages` are 
+// provided as command line operands.The included record `person` represents 
+// the options given via the command line.
+// The `main` function may return an `error` or `()`.
+public function main(int repeats, string[] messages, *Person person)
```
warunalakshitha, line 18: Some example suggestion


main(string subjectName, string[] studentNames, int[] marks, *ResultOption resultOption)


public type ResultOption record {|
    int passMark;
    boolean? printAverage;	
|};


bal runa main.bal -- --passMArk 50 --printAverage -- Science Tom Alice John 70 30 80

output

Tom    P
Alice  F
John   P 

Average 60

sameerajayasoma: With the new command-line args model, you can have at most one operand parameter of type T[]. This example is not valid. Am I missing something? :)

sameerajayasoma (changes requested):
Since we are rethinking these examples, can we find a better real-world example?  
How about implementing the grep command in Ballerina? 

```grep [options] <pattern> <filename>...```

e.g.,
```grep -i "foo" bar.txt bazz.txt```

**Comment by sameerajayasoma** [comment]
Perhaps we should keep `The Main Function` example as simple as possible. 
```
public function main(string[] args) {
}
```

We can introduce a different set of examples to cover command-line args processing.

File: examples/the-main-function/the_main_function.bal (hunk 1 above)
Comment (riyafa) [reply examples/the-main-function/the_main_function.bal:18]: According to spec we cannot have `string[]` and `string?` or `string v = "d"`, but it allows having `string`
Quoting from the spec:

> if there is an operand parameter of type O[], then there must be no operand parameters of type O? and O x = d

Comment (warunalakshitha) [reply examples/the-main-function/the_main_function.bal:18]: ATM multiple operand parameters with type T[] also allowed and works fine. @riyafa seems spec deviation and needs a fix.

Comment (riyafa) [reply examples/the-main-function/the_main_function.bal:18]: @sameerajayasoma I think spec allows having `string[]` and `int[]` arrays. But does not allow having two `string[]` arrays

Comment (warunalakshitha) [reply examples/the-main-function/the_main_function.bal:18]: AFAIU we can not have int[] and string[] which are subtypes of basic types and cannot have multiple operand parameters of type T[].

Comment (riyafa) [reply examples/the-main-function/the_main_function.bal:18]: @sameerajayasoma can you please clarify, if the spec does or does not allow arrays of different types? I have been under the impression, it does not allow arrays of same type.

**Comment by riyafa** [comment]
> Since we are rethinking these examples, can we find a better real-world example?
> How about implementing the grep command in Ballerina?
> 
> `grep [options] <pattern> <filename>...`
> 
> e.g.,
> `grep -i "foo" bar.txt bazz.txt`

I have also been considering things like the `ls` command, but dismissed the idea because we might complicate the example because of having to read from files (which is in itself a separate example).  But I guess these are the most important use cases of cli parsing so I will try to write such an example.

**Comment by sameerajayasoma** [comment]
> I have also been considering things like the `ls` command, but dismissed the idea because we might complicate the example because of having to read from files (which is in itself a separate example). But I guess these are the most important use cases of cli parsing so I will try to write such an example.

We can simply show how you can design a command-line interface using Ballerina. We don't have to show to you read files etc. We can simply print the operands and options.

**Comment by riyafa** [comment]
> Perhaps we should keep `The Main Function` example as simple as possible.
> 
> ```
> public function main(string[] args) {
> }
> ```
> 
> We can introduce a different set of examples to cover command-line args processing.

There is a simple example for `The Main Function` which is the `Hello World Main` example. 
This example has been there to show the possible arguments to the main function. 
Since there's already a simple example for the main function, shall we remove this example and create a new one for command line parsing under a different heading? (This is under the `functions` heading)

File: examples/the-main-function/the_main_function.bal [hunk 2]
```diff
@@ -1,26 +1,49 @@
...
-// The first parameter `name` is a required parameter, while the second parameter `age` is a defaultable
-// parameter with the default value `18`. The third parameter `year` is also a defaultable parameter.
-// The rest parameter `modules` represents the additional arguments.
-// The `main` function may return an `error` or `()`.
-public function main(string name,
-                     int age = 18,
-                     string year = "Freshman",
-                     string... modules)
-                  returns error? {
+// These options are some of those found in the `grep` command line
```
praneesha, line 3:
```suggestion
// These options are some of those found in the [`grep` command line 
```

examples/the-main-function/the_main_function.bal
```diff
@@ -1,26 +1,49 @@
...
-// parameter with the default value `18`. The third parameter `year` is also a defaultable parameter.
-// The rest parameter `modules` represents the additional arguments.
-// The `main` function may return an `error` or `()`.
-public function main(string name,
-                     int age = 18,
-                     string year = "Freshman",
-                     string... modules)
-                  returns error? {
+// These options are some of those found in the `grep` command line 
+// utility in [here](https://www.man7.org/linux/man-pages/man1/grep.1.html).
```
praneesha, line 4:
```suggestion
// utility](https://www.man7.org/linux/man-pages/man1/grep.1.html).
```

examples/the-main-function/the_main_function.bal
```diff
@@ -1,26 +1,49 @@
...
+    boolean 'line\-number;
+    boolean recursive;
+    string? file;
+|};
 
-    // Return an error if the name is invalid.
-    if (name.length() < 5) {
-        error e = error("InvalidName", message = "invalid length");
-        return e;
+// The main function accepts a `pattern` to search for in the `fileName`
```
praneesha, line 14:
```suggestion
// The main function accepts a `pattern` to search for the `fileName`
```

examples/the-main-function/the_main_function.bal
```diff
@@ -1,26 +1,49 @@
...
+// and the arguments for the linux/unix `grep` command. This program 
+// processes the command line arguments and prints their value with an
```
praneesha, line 16:
```suggestion
// processes the command line arguments and prints their values with an 
```

examples/the-main-function/the_main_function.bal
```diff
@@ -1,26 +1,49 @@
...
-
-    if (modules.length() > 0) {
-        info += ", Module(s): " + modules.toString();
+    if (grepOptions.'line\-number) {
+        io:println("Precede each output line by its line number");
+    }
+    if (grepOptions.recursive) {
+        io:println("Recursively search subdirectories listed");
+    }
+    // Handles optional `file` option.
```
praneesha, line 39:
```suggestion
    // Handles the optional `file` option.
```

examples/the-main-function/the_main_function.out
```diff
@@ -1,31 +1,16 @@
 # To run this sample, navigate to the directory that contains the
 # `.bal` file, and execute the `bal run` command below.
 
-# Use the ballerina `run` command to invoke the `main` function specifying `Alice`
-# as the string argument for `name`. `18` would be set as the value for
-# `age` and `Freshman` would be set as the value for `year`.
-bal run the_main_function.bal -- Alice
-Name: Alice, Age: 18, Year: Freshman
+# Use the ballerina `run` command to invoke the `main` function 
+# specifying the value `3` as the`after-context`,
```
praneesha, line 5:
```suggestion
# specifying the value `3` as the`after-context` and
```

examples/the-main-function/the_main_function.out
```diff
@@ -1,31 +1,16 @@
...
 
-# Use the ballerina `run` command to invoke the `main` function specifying `Alice`
-# as the string argument for `name`. `18` would be set as the value for
-# `age` and `Freshman` would be set as the value for `year`.
-bal run the_main_function.bal -- Alice
-Name: Alice, Age: 18, Year: Freshman
+# Use the ballerina `run` command to invoke the `main` function 
+# specifying the value `3` as the`after-context`, 
+# the value `4` as the `before-context`.
+# The values of the `ignore-case`, `recursive` and `line-number` options
```
praneesha, line 7:
```suggestion
# The values of the `ignore-case`, `recursive`, and `line-number` options 
```

examples/the-main-function/the_main_function.out
```diff
@@ -1,31 +1,16 @@
...
-# Use the ballerina `run` command to invoke the `main` function specifying `Alice`
-# as the string argument for `name`. `18` would be set as the value for
-# `age` and `Freshman` would be set as the value for `year`.
-bal run the_main_function.bal -- Alice
-Name: Alice, Age: 18, Year: Freshman
+# Use the ballerina `run` command to invoke the `main` function 
+# specifying the value `3` as the`after-context`, 
+# the value `4` as the `before-context`.
+# The values of the `ignore-case`, `recursive` and `line-number` options 
+# can be set to `true` by specifying them explicitly .
```
praneesha, line 8:
```suggestion
# can be set to `true` by specifying them explicitly.
```

examples/the-main-function/the_main_function.out
```diff
@@ -1,31 +1,16 @@
...
+# `file` is an optional option argument which is the the name of `file` with newline separated patterns to search for.
```
praneesha, line 9:
```suggestion
# `file` is an optional option argument, which is the name of `file` with newline-separated patterns to search for.
```

examples/the-main-function/the_main_function.out
```diff
@@ -1,31 +1,16 @@
...
+# The operand arguments for `pattern` and `fileName` can be specified along 
+# with the options, but they have to be specified in the same order.
```
praneesha, line 11:
```suggestion
# with the options but they have to be specified in the same order. 
```

examples/the-main-function/the_main_function.out
```diff
@@ -1,31 +1,16 @@
...
-Name: Alice, Age: 18, Year: Freshman
+# Use the ballerina `run` command to invoke the `main` function 
+# specifying the value `3` as the`after-context`, 
+# the value `4` as the `before-context`.
+# The values of the `ignore-case`, `recursive` and `line-number` options 
+# can be set to `true` by specifying them explicitly .
+# `file` is an optional option argument which is the the name of `file` with newline separated patterns to search for.
+# The operand arguments for `pattern` and `fileName` can be specified along 
+# with the options, but they have to be specified in the same order. 
+# Here `pattern` is specified as `context` and `fileName` is specified
```
praneesha, line 12:
```suggestion
# Here, `pattern` is specified as `context` and `fileName` is specified
```

examples/the-main-function/the_main_function.bal
riyafa: I am not sure I understand this change. Is this necessary?

PR #1706, opened 2021-05-18
Refactor mysql call stored procedure BBE

kalaiyarasiganeshalingam:
## Purpose
$Subject

### DIFF
--- a/examples/mysql-call-stored-procedures/afterExample.out
+++ b/examples/mysql-call-stored-procedures/afterExample.out
@@ -1,8 +1,17 @@
-# To create the `Ballerina` project, execute the `bal run` command below and copy this code to that project
+# To create the `Ballerina` project, execute the `bal run` command below
+# and copy this code to that project
 bal new afterExample
+
+# To define the MySQL database driver JAR as a dependency, change the path of
+# the MYSQL driver and add it in the `Ballerina.toml`
+[[platform.java11.dependency]]
+artifactId = "mysql-connector-java"
+version = "VERSIONN"
+path = "/path/to/mysql-connector-java.jar"
+groupId = "mysql"
 
 # To run this, execute the `bal run` command below.
 bal run afterExample
 
-Starting to drop the tables...
-Drop the tables successfully done!
+Starts to clean bbe tables...
+Dropped table successfully!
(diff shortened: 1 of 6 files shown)


examples/mysql-call-stored-procedures/mysql_call_stored_procedures.bal
```diff
@@ -119,68 +58,44 @@ type Student record {
...
-            sql:Error? e = studentStream.forEach(function(Student student) {
-                                io:println("Student details: ", student);
-                           });
-        } else {
-            io:println("Empty result is returned from the `GetStudents`.");
-        }
-        checkpanic retCall.close();
+    io:println("Call stored procedure `InsertStudent` is successful.");
+
+    // Process the returned result stream.
```
praneesha, line 71:
```suggestion
    // Processes the returned result stream.
```

kalaiyarasiganeshalingam:
> Do not use separate function, move them all here

If this sample demonstrates one function, we can write under the main method. But this sample has three functions. So, If those move under the main method, that method will become multiple lines coding. Therefore, I thought if dividing those into separate methods, that is more clear.  WDYT?

niveathika:
> > Do not use separate function, move them all here
> 
> If this sample demonstrates one function, we can write under the main method. But this sample has three functions. So, If those move under the main method, that method will become multiple lines coding. Therefore, I thought if dividing those into separate methods, that is more clear. WDYT?

The thing is, we have got feedback on the use of functions it's hard to understand the flow since the main method is the last one in the bal file. Though it's the best programming practice when you read the be, it's hard to understand. How about we combine all 3 scenarios into one procedure call and explain. Or I am +1 to show only one scenario too, the others we can explain in the API docs

examples/mysql-call-stored-procedures/afterExample.out
```diff
@@ -0,0 +1,17 @@
+# To create the `Ballerina` project, execute the `bal run` command below
+# and copy this code to that project
+bal new afterExample
+
+# To define the MySQL database driver JAR as a dependency, change the path of
```
kalaiyarasiganeshalingam, line 5: @daneshk @niveathika I have added this info here. Are you accept this?

praneesha (approved):

PR #1727, opened 2021-05-20
Add new lang BBEs

lasinicl:
## Purpose
$title for BBE revamping effort


new-examples/raw-templates/raw_templates.bal
```diff
@@ -1 +1,33 @@
+import ballerina/io;
+import ballerinax/java.jdbc;
+import ballerina/sql;
 
+jdbc:Client dbClient = check new (url = "jdbc:h2:file:./master/orderdb",
```
hasithaa, line 5: @lasinicl We didn't use other std libs so far. Shall come up with an example without an std lib. @sameerajayasoma wdyt ?

hasithaa:
I will merge this for now.

PR #1846, opened 2021-06-01
Fix compilation issues with wait in BBEs

warunalakshitha:
## Purpose
$Subject


examples/named-workers-and-futures/named_workers_and_futures.bal
```diff
@@ -22,10 +22,10 @@ function startInt(FuncInt f) returns future<int> {
 
 public function main() {
     future<int> a = demo();
-    int b = wait a;
+    int b = checkpanic wait a;
```
niveathika, line 25: Cant we use check and return the error?

niveathika (approved):

PR #1882, opened 2021-06-02
Update nats, stan and rabbitmq BBEs

aashikam:
## Purpose
$title


examples/nats-basic-pub-sub/subscriber.out
```diff
@@ -1,5 +1,3 @@
-# To run this sample, navigate to the directory that contains the
-# `.bal` file, and execute the `bal run` command below.
 bal run subscriber.bal
```
ThisaruGuruge, line 2: Do we need this new line? (I get that output verification does not work for the messaging ones, but if we had verification, this will fail I guess).

ThisaruGuruge (approved):

PR #1886, opened 2021-06-02
Update instructions of security BBEs

ldclakmal:
## Purpose
$subject

Removed the instruction:
```
# To start the service, navigate to the directory that contains the
# `.bal` file and use the `bal run` command below.
```
Related to: Related to https://github.com/ballerina-platform/ballerina-standard-library/issues/584

### DIFF
--- a/examples/graphql-service-basic-auth-ldap-user-store/graphql_service_basic_auth_ldap_user_store.server.out
+++ b/examples/graphql-service-basic-auth-ldap-user-store/graphql_service_basic_auth_ldap_user_store.server.out
@@ -1,5 +1,3 @@
-# To run this sample, navigate to the directory that contains the `.bal` file,
-# and execute the `bal run` command below.
 # (You may need to change the certificate file path and private key file path.)
 bal run graphql_service_basic_auth_ldap_user_store.bal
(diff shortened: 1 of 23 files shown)


examples/graphql-service-basic-auth-ldap-user-store/graphql_service_basic_auth_ldap_user_store.server.out
```diff
@@ -1,5 +1,3 @@
-# To run this sample, navigate to the directory that contains the `.bal` file,
-# and execute the `bal run` command below.
 # (You may need to change the certificate file path and private key file path.)
```
ThisaruGuruge, line 1: Do we need these brackets anymore? Since this is the only instruction here, we can remove the bracket, right? Check the other place as well.

ldclakmal: Agree. Fixed with https://github.com/ballerina-platform/ballerina-distribution/pull/1886/commits/f3fca78c22aaa78dbea2491491bc265438df6fce

ThisaruGuruge (approved):

PR #1990, opened 2021-06-21
Improve Table Syntax BBE description

lasinicl:
## Purpose
$title

### DIFF
--- a/examples/table-syntax/table_syntax.description
+++ b/examples/table-syntax/table_syntax.description
@@ -1,15 +1,15 @@
 // A `table` is a collection of records. Each `record` represents a row of the `table`.
-// A `table` is plain data if and only if its rows are plain data. Table maintains an invariant
+// A `table` is plain data if and only if its rows are plain data. A `table` maintains an invariant
 // that each row is uniquely identified by a key. Each row’s key is stored in fields, which
 // must be immutable. <br/><br/>
 // Compared to maps,
 // <ul>
-// <li>key is part of the value, rather than separate</li>
-// <li>type of the key is not restricted to `string`</li>
-// <li>order of members is preserved</li>
+// <li>key is part of the value rather than being separate.</li>
+// <li>The type of the key is not restricted to `string`.</li>
+// <li>The order of the members is preserved.</li>
 // </ul>
 // <br/><br/>
 // <p>A `record` field can be declared as `readonly`. A value cannot be assigned to such a field
-// after the record is created. `table` type gives the type of the row and name of the key field.
-// `table constructor expression` looks like an `array constructor`. `foreach` statement will
-// iterate over a table's rows in order. Use `t[k]` to access a row using its key.</p>
+// after the record is created. The `table` type gives the type of the row and the name of the key field.
+// The `table constructor expression` looks like an `array constructor`. The `foreach` statement will
+// iterate over a table's rows in their order. Use `t[k]` to access a row using its key.</p>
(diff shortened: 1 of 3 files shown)


examples/table-syntax/table_syntax.description
```diff
@@ -1,4 +1,15 @@
...
+// <ul>
+// <li>key is part of the value, rather than separate</li>
+// <li>type of the key is not restricted to `string`</li>
+// <li>order of members is preserved</li>
+// </ul>
+// <br/><br/>
+// <p>A `record` field can be declared as `readonly`. A value cannot be assigned to such a field
+// after the record is created. `table` type gives the type of the row and name of the key field.
+// `table constructor expression` looks like an `array constructor`. `foreach` statement will
+// iterate over a table's rows in order. Use `t[k]` to access a row using its key.</p>
```
pcnfernando, line 15: I think we can avoid re-stating that the table maintains the order. wdyt?

lasinicl: In the places where the order is mentioned, 
the first one talks about how tables are different from maps and 
the other one talks about how a `foreach` iterates a table. 
I don't feel like it's a repetition of the same thing.

examples/table-syntax/table_syntax.description
```diff
@@ -1,4 +1,15 @@
-// A `record` field can be declared as `readonly`: cannot assign to the field after the record is created.
-// `table` type gives type of row and name of key field. `table constructor expression` looks like an
-// `array constructor`. `foreach` statement will iterate over a table's rows in order. Use `t[k]`
-// to access a row using its key.
+// A `table` is a collection of records. Each `record` represents a row of the `table`.
+// A `table` is plain data if and only if its rows are plain data. Table maintains an invariant
```
praneesha, line 2:
```suggestion
// A `table` is plain data if and only if its rows are plain data. A `table` maintains an invariant
```

examples/table-syntax/table_syntax.description
```diff
@@ -1,4 +1,15 @@
...
+// that each row is uniquely identified by a key. Each row’s key is stored in fields, which
+// must be immutable. <br/><br/>
+// Compared to maps,
+// <ul>
+// <li>key is part of the value, rather than separate</li>
```
praneesha, line 7:
```suggestion
// <li>key is part of the value rather than being separate.</li>
```

examples/table-syntax/table_syntax.description
```diff
@@ -1,4 +1,15 @@
...
+// <li>type of the key is not restricted to `string`</li>
```
praneesha, line 8:
```suggestion
// <li>The type of the key is not restricted to `string`.</li>
```

examples/table-syntax/table_syntax.description
```diff
@@ -1,4 +1,15 @@
...
+// <li>order of members is preserved</li>
```
praneesha, line 9:
```suggestion
// <li>The order of the members is preserved.</li>
```

examples/table-syntax/table_syntax.description
```diff
@@ -1,4 +1,15 @@
...
+// </ul>
+// <br/><br/>
+// <p>A `record` field can be declared as `readonly`. A value cannot be assigned to such a field
+// after the record is created. `table` type gives the type of the row and name of the key field.
```
praneesha, line 13:
```suggestion
// after the record is created. The `table` type gives the type of the row and the name of the key field.
```

examples/table-syntax/table_syntax.description
```diff
@@ -1,4 +1,15 @@
...
+// `table constructor expression` looks like an `array constructor`. `foreach` statement will
```
praneesha, line 14:
```suggestion
// The `table constructor expression` looks like an `array constructor`. The `foreach` statement will
```

examples/table-syntax/table_syntax.description
praneesha:
```suggestion
// iterate over a table's rows in their order. Use `t[k]` to access a row using its key.</p>
```

praneesha (approved):

pcnfernando (approved):

PR #1997, opened 2021-06-22
 Add build-time-tests module

azinneera:

### DIFF
--- a/build-time-tests/build.gradle
+++ b/build-time-tests/build.gradle
@@ -50,7 +50,6 @@ task buildSamples {
     }
 }
 
-
 task processData(type: JavaExec) {
     dependsOn buildSamples
     classpath = sourceSets.main.runtimeClasspath
--- a/build-time-tests/build.gradle
+++ b/build-time-tests/build.gradle
@@ -1,0 +1,59 @@
+/*
+ ~ * Copyright (c) 2021, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ ~ *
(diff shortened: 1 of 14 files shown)


build-time-tests/build.gradle
```diff
@@ -0,0 +1,59 @@
...
+    dependsOn configurations.ballerinaDistribution
+    dependsOn (":ballerina:unzipDistForTests")
+    doLast {
+        exec {
+            workingDir "${buildDir}/samples"
+            commandLine 'sh', '-c', "bash build-samples.sh $distPath"
+        }
+    }
+}
+
```
keizer619, line 52:
```suggestion
```

build-time-tests/build.gradle
```diff
@@ -0,0 +1,59 @@
...
+task buildSamples {
+    dependsOn copySamples
+    dependsOn configurations.ballerinaDistribution
+    dependsOn (":ballerina:unzipDistForTests")
+    doLast {
+        exec {
+            workingDir "${buildDir}/samples"
+            commandLine 'sh', '-c', "bash build-samples.sh $distPath"
```
keizer619, line 48: Apparently this is only being used in ubuntu workflow. Do we need to consider windows user machine scenarios?

pramodya1994 (approved):

build-time-tests/build.gradle
azinneera: Not necessary as of now. We will add it later if needed.

PR #2082, opened 2021-08-03
Fix resolving bal path in windows build

NipunaMadhushan:
## Purpose
Enable tests in windows build


ballerina-test/src/test/java/org/ballerinalang/distribution/utils/TestUtils.java
```diff
@@ -81,7 +81,12 @@ private static void logOutput(InputStream inputStream) throws IOException {
     public static boolean executeBuild(String distributionName, Path sourceDirectory, List<String> args) throws
             IOException, InterruptedException {
         args.add(0, "build");
-        args.add(0, TEST_DISTRIBUTION_PATH.resolve(distributionName).resolve("bin").resolve("bal").toString());
+        if (System.getProperty("os.name").startsWith("Windows")) {
```
keizer619, line 84: Shall use this check and assign file name "bal or bal.bat" to a variable and use that variable in onestatement

PR #2117, opened 2021-08-20
Add Manu's automated stdlib build scripts

azinneera:


hevayo (commented):
Shall we add a read me too

hevayo (approved):

ThisaruGuruge:
I have a better script for this. 

The issues I find in this script are:
1. The module list can be changed and we have to manually change it.
2. We may have to execute different commands at different times, (skip tests, or not skip tests, update lang version in each repo, etc.)

azinneera:
We don't need this PR as there is a better script for this written by @ThisaruGuruge.

PR #2180, opened 2021-09-28
Add MacOS pkg installer for daily-build & release workflows

NipunaMadhushan:
## Purpose
Remove jenkins from the release pipeline.

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -140,3 +140,35 @@ jobs:
         run: |
           curl -fsSL https://github.com/github/hub/raw/master/script/get | bash -s 2.14.1
           bin/hub pull-request -m "[Automated] Sync master after "$VERSION" release"
+
+    outputs:
+      project-version: ${{ steps.version-set.outputs.version }}
+      upload-asset-url: ${{ steps.create_release.outputs.upload_url }}
+      release-version: ${{ steps.version-set.outputs.taggedVersion }}
+
+  macos-installer-build:
+
+    needs: publish-release
+    runs-on: macos-latest
+
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-release.yml
```diff
@@ -140,3 +140,52 @@ jobs:
...
+
+  macos-build:
+
+    needs: publish-release
+    runs-on: macos-latest
+
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v2
+      - name: Set up JDK 11
```
keizer619, line 157: @NipunaMadhushan We should be able to remove this as installer generation does not need jdk11

.github/workflows/publish-release.yml
```diff
@@ -140,3 +140,52 @@ jobs:
...
+      - name: Upload MacOS pkg Installer
+        uses: actions/upload-release-asset@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.BALLERINA_BOT_TOKEN }}
+        with:
+          upload_url: ${{ needs.publish-release.outputs.upload-asset-url }}
+          asset_name: ballerina-macos-installer-x64-${{ needs.publish-release.outputs.project-version }}.pkg
+          asset_path: installers/mac/target/pkg/ballerina-macos-installer-x64-${{ needs.publish-release.outputs.project-version }}.pkg
+          asset_content_type: application/octet-stream
+      - name: Upload MacOS zip artifacts
```
keizer619, line 183: @NipunaMadhushan Do we need to upload zip again as this zip has already uploaded by ubuntu job as it was built there ?

.github/workflows/publish-release.yml
NipunaMadhushan: I will remove this.

keizer619 (approved):

PR #2224, opened 2021-10-10
Fix the load balancer BBE for isolated

MaryamZi:
## Purpose
$title. By making the module-level client final so that the method is inferred to be isolated.


MaryamZi:
Closing since this will be introduced with https://github.com/ballerina-platform/ballerina-distribution/pull/2225.

PR #2297, opened 2021-11-02
Ignore npm CVEs

praveennadarajah:
## Purpose
$subject


keizer619 (approved):

keizer619:
I analyzed above and we can keep it ignored. This happens inside bala we have a `package.json`. There ballerina mime version is `2.0.1`. When image is being scanned it gets confused and consider `package.json` as a npm module and asks us to update to `2.0.3`.

PR #2377, opened 2021-11-29
Update the display text

keizer619:
## Purpose
Update the display text


NipunaRanasinghe (commented):
Dont we need to change `swan-lake-latest-version` as well?

NipunaRanasinghe (approved):

Issue #2424, opened 2021-12-14
Create a workflow to run installer tests for signed pkg installer


NipunaMadhushan:
**Description:**
Create a workflow to run installer tests for signed pkg installer in the release.

**Describe your task(s)**
Currently the pkg installer created from the release workflow is not signed. Therefore it needs to be signed manually and replace it. Then after that the installer tests should be run against the signed pkg installer again and a workflow is needed to run the tests.

PR #2426, opened 2021-12-14
Create a workflow to test signed pkg installer

NipunaMadhushan:
## Purpose
Test signed pkg installer by running installer tests

## Goals
Test release artifacts after the signing process

## Fixes
https://github.com/ballerina-platform/ballerina-distribution/issues/2424

### DIFF
--- a/.github/workflows/test-installers.yml
+++ b/.github/workflows/test-installers.yml
@@ -1,0 +1,39 @@
+name: Test Installers
+
+on:
+  workflow_dispatch:
+    inputs:
+      releaseVersion:
+        description: 'Version of the release. e.g., swan-lake-beta1'
+        required: true
+      preReleaseSuffix:
+        description: 'The text that will be suffixed to the Git tag. e.g., rc1'
+        required: false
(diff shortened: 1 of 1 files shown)


.github/workflows/test-installers.yml
```diff
@@ -0,0 +1,26 @@
+name: Test Installers
+
+on:
+  workflow_dispatch:
+    inputs:
+      releaseVersion:
+        description: 'Version of the release. e.g., swan-lake-beta1'
+        required: true
+      taggedVersion:
+        description: 'Tagged version of the release. e.g., swan-lake-beta1-rc1, swan-lake-beta1'
```
keizer619, line 10: @NipunaMadhushan Shall we only expect only the pre release part as the input (eg: "-rc1" or "rc1"). Then we can construct entire thing with "releaseVersion" input

NipunaMadhushan: @keizer619 I will address the suggestions.

keizer619 (approved):

PR #2501, opened 2022-01-21
Add AsyncAPI module

HasithaAthukorala:
Add AsyncAPI module to Ballerina Distribution


pubudu91:
Let's not merge this if this is not required for the GA

PR #2514, opened 2022-01-24
Fix cache, file and task BBE

kalaiyarasiganeshalingam:
## Purpose
$Subject

### DIFF
--- a/examples/files/files.out
+++ b/examples/files/files.out
@@ -1,6 +1,6 @@
 bal run files.bal
 The bar.txt file created successfully.
-Is the bar.txt file exists: true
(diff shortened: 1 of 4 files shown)


examples/files/files.out
```diff
@@ -1,11 +1,12 @@
 bal run files.bal
 The bar.txt file created successfully.
-Is the bar.txt file exists: true
+Is the bar.txt file exist: true
```
pubudu91, line 3:
```suggestion
Does the bar.txt file exist: true
```

daneshk (approved):

PR #2542, opened 2022-01-29
Fix bbe descriptions and formatting

lochana-chathura:
## Purpose
$subject.

### DIFF
--- a/examples/const-and-final/const_and_final.description
+++ b/examples/const-and-final/const_and_final.description
@@ -1,3 +1,3 @@
-// `const` means immutable and known at compile-time. Its type is singleton: set containing single value.
-// A `variable` or a `class` field can be declared as `final`, meaning it cannot be assigned to, after
+// `const` means immutable and known at compile-time. Its type is singleton: a set containing a single value.
+// A `variable` or a `class` field can be declared as `final`, meaning it cannot be assigned after
 // it has been initialized.
--- a/examples/const-and-final/const_and_final.description
+++ b/examples/const-and-final/const_and_final.description
@@ -1,3 +1,3 @@
 // `const` means immutable and known at compile-time. Its type is singleton: a set containing a single value.
-// A `variable` or a `class` field can be declared as `final`, which means it cannot be assigned after
+// A variable or a class field can be declared as `final`, which means it cannot be assigned after
 // it has been initialized.
--- a/examples/const-and-final/const_and_final.description
+++ b/examples/const-and-final/const_and_final.description
@@ -1,3 +1,3 @@
 // `const` means immutable and known at compile-time. Its type is singleton: a set containing a single value.
-// A `variable` or a `class` field can be declared as `final`, meaning it cannot be assigned after
+// A `variable` or a `class` field can be declared as `final`, which means it cannot be assigned after
 // it has been initialized.
--- a/examples/enumerations/enumerations.bal
+++ b/examples/enumerations/enumerations.bal
@@ -10,7 +10,9 @@ import ballerina/io;
 //
 // `type Color RED|GREEN|BLUE;`
 enum Color {
-    RED, GREEN, BLUE
+    RED,
+    GREEN,
+    BLUE
 }
 
 // An `enum` member can explicitly specify an associated expression.
--- a/examples/identity/identity.bal
+++ b/examples/identity/identity.bal
@@ -9,21 +9,32 @@ public function main() {
     MyClass obj2 = new MyClass();
     // `b1` will be true.
     boolean b1 = (obj1 === obj1);
+
+    io:println(b1);
+
     // `b2` will be false.
     boolean b2 = (obj1 === obj2);
+
+    io:println(b2);
+
     // `b3` will be true.
     boolean b3 = ([1, 2, 3] == [1, 2, 3]);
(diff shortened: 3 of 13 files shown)


examples/enumerations/enumerations.bal
```diff
@@ -10,7 +10,9 @@ import ballerina/io;
 //
 // `type Color RED|GREEN|BLUE;`
 enum Color {
-    RED, GREEN, BLUE
+    RED,
+    GREEN,
+    BLUE
 }
```
lochana-chathura, line 17: Ballerina formatter does not allow writing enums with members in-line. I think we need to allow it. Created ballerina-platform/ballerina-lang#34896. Anyway, changed the bal according to the current formatting.

examples/identity/identity.bal
```diff
@@ -5,17 +5,17 @@ class MyClass {
 public function main() {
     MyClass obj1 = new MyClass();
     MyClass obj2 = new MyClass();
-    // `b1` will be true.
+    // The result will be true.
     boolean _ = (obj1 === obj1);
```
lochana-chathura, line 9: Instead of saying `The result will be false.` Maybe we can assign a variable as previously, say b1, and then io:println() all of them at the end? This will also make sure the bbe output is correct.

gimantha: +1, we do have "run in playground" button in BBEs. So if we are not printing anything, users wont see anything.

examples/const-and-final/const_and_final.description
```diff
@@ -1,3 +1,3 @@
-// `const` means immutable and known at compile-time. Its type is singleton: set containing single value.
-// A `variable` or a `class` field can be declared as `final`, meaning it cannot be assigned to, after
+// `const` means immutable and known at compile-time. Its type is singleton: a set containing a single value.
+// A `variable` or a `class` field can be declared as `final`, meaning it cannot be assigned after
```
praneesha, line 2:
```suggestion
// A `variable` or a `class` field can be declared as `final`, which means it cannot be assigned after
```

examples/identity/identity.description
```diff
@@ -1,2 +1,2 @@
-// `===` and `!==` operators test for identity. Identical for mutable basic types means stored at the same address.
+// `===` and `!==` operators test for identity. The identity is identical for mutable basic types means they are stored at the same address.
```
gimantha, line 1:
```suggestion
// `===` and `!==` operators test for identity. Identical mutable basic types means they are stored at the same address.
```

examples/identity/identity.bal
```diff
@@ -6,16 +8,22 @@ public function main() {
...
     // `b6` will be false.
-    boolean _ = (-0.0 === +0.0);
+    boolean b6 = (-0.0 === +0.0);
 
+    io:println(b1);
+    io:println(b2);
+    io:println(b3);
+    io:println(b4);
+    io:println(b5);
+    io:println(b6);
```
MaryamZi, line 28: Shall we print these immediately after defining the variable? Easier to follow IMO.

examples/identity/identity.description
```diff
@@ -1,2 +1,2 @@
-// `===` and `!==` operators test for identity. Identical for mutable basic types means stored at the same address.
+// `===` and `!==` operators test for identity. Identical mutable basic types means that they are stored at the same address.
```
MaryamZi, line 1:
```suggestion
// `===` and `!==` operators test for identity. Values of mutable basic types are identical if and only if they are stored at the same address.
```

examples/identity/identity.description
```diff
@@ -1,2 +1,2 @@
-// `===` and `!==` operators test for identity. Identical for mutable basic types means stored at the same address.
+// `===` and `!==` operators test for identity. Values of mutable basic types are identical if and only if they are stored at the same address.
```
praneesha, line 1:
```suggestion
// `===` and `!==` operators test for identity. Values of mutable basic types are identical if and only if they are stored in the same address.
```

examples/immutability/immutability.bal
```diff
@@ -13,7 +13,7 @@ public function main() {
 
         grade: 12,
         name: "John",
-        // The applicable contextually-expected type for `marks`.
+        // The applicable contextually-expected type for `marks`
         // is now `map<int> & readonly`. Thus the value for `marks`
```
praneesha, line 17:
```suggestion
        // is now `map<int> & readonly`. Thus, the value for `marks`
```

examples/const-and-final/const_and_final.description
```diff
@@ -1,3 +1,3 @@
-// `const` means immutable and known at compile-time. Its type is singleton: set containing single value.
-// A `variable` or a `class` field can be declared as `final`, meaning it cannot be assigned to, after
+// `const` means immutable and known at compile-time. Its type is singleton: a set containing a single value.
+// A `variable` or a `class` field can be declared as `final`, which means it cannot be assigned after
```
pubudu91, line 2:
```suggestion
// A variable or a class field can be declared as `final`, which means it cannot be assigned after
```
Since,
- "variable" is not a keyword
- "class" is not used in its keyword context

pubudu91 (approved):

PR #2546, opened 2022-01-29
Minor update to limit clause description

KRVPerera:
## Purpose
```diff
-The `limit` clause limits the number of results from the earlier clauses.
+The `limit` clause limits the number of results from the query.
```

### DIFF
--- a/examples/limit-clause/limit_clause.description
+++ b/examples/limit-clause/limit_clause.description
@@ -1,1 +1,1 @@
-// The `limit` clause limits the number of results from the earlier clauses.
+// The `limit` clause limits the number of results from the query.


praneesha (approved):

examples/limit-clause/limit_clause.description
```diff
@@ -1 +1 @@
-// The `limit` clause limits the number of results from the earlier clauses.
+// The `limit` clause limits the number of results from the query.
```
MaryamZi, line 1: Is this change correct? A query expression can have multiple limit clauses? @gimantha

KRVPerera: limits the frames (results) from the query.

MaryamZi: Limits them for the next clause, right? Not necessarily for the entire query expression?

KRVPerera: I got it. There can be multiple limit clauses.

PR #2547, opened 2022-01-29
Update querying tables BBE

KRVPerera:
## Purpose
Return type of query expression on a table can be a table too

### DIFF
--- a/examples/querying-tables/querying_tables.bal
+++ b/examples/querying-tables/querying_tables.bal
@@ -16,7 +16,8 @@ public function main() {
         {id: 9, firstName: "Cassie", lastName: "Smith", salary: 250}
     ];
 
-    // `from` clause iterates `employees` `table` and creates a `int[]`.
+    // The `from` clause iterates `employees` table.
+    // The query expression creates an integer array based on the contextually-expected type.
     int[] salaries = from var {salary} in employees
                      select salary;
 
--- a/examples/querying-tables/querying_tables.bal
+++ b/examples/querying-tables/querying_tables.bal
@@ -16,7 +16,7 @@ public function main() {
         {id: 9, firstName: "Cassie", lastName: "Smith", salary: 250}
     ];
 
-    // The `from` clause iterates `employees` table.
+    // The `from` clause iterates the `employees` table.
     // Since the contextually-expected type for the query expression is `int[]` evaluation of the query expression will result in an integer array.
     int[] salaries = from var {salary} in employees
                      select salary;
--- a/examples/querying-tables/querying_tables.bal
+++ b/examples/querying-tables/querying_tables.bal
@@ -29,6 +29,6 @@ public function main() {
                      where emp.lastName == "Smith" select emp;
 
     foreach Employee emp in smiths {
-        io:println(emp.firstName + " " + emp.lastName);
+        io:println(emp.firstName, " ", emp.lastName);
     }
 }
--- a/examples/querying-tables/querying_tables.bal
+++ b/examples/querying-tables/querying_tables.bal
@@ -17,7 +17,7 @@ public function main() {
     ];
 
     // The `from` clause iterates `employees` table.
-    // The query expression creates an integer `array` based on the contextually-expected type.
+    // Since the contextually-expected type for the query expression is `int[]` evaluation of the query expression will result in an integer array.
     int[] salaries = from var {salary} in employees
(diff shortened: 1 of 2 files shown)


examples/querying-tables/querying_tables.bal
```diff
@@ -10,14 +10,24 @@ type Employee record {|
...
-        {id: 2, firstName: "Fred", lastName: "Bloggs", salary: 200}
+        {id: 2, firstName: "Jane", lastName: "Smith", salary: 100},
+        {id: 4, firstName: "Fred", lastName: "Bloggs", salary: 200},
+        {id: 7, firstName: "Bobby", lastName: "Clark", salary: 200},
+        {id: 9, firstName: "Cassie", lastName: "Smith", salary: 200}
     ];
 
-    // `from` clause iterates `employees` `table`.
-    // The contextually-expected type of the `query expression` is an `int[]`.
+    // `from` clause iterates `employees` `table` and creates a `int[]`.
```
MaryamZi, line 19:
```suggestion
    // The `from` clause iterates `employees` table.
    // The query expression creates an integer array based on the contextually-expected type.
```

examples/querying-tables/querying_tables.bal
```diff
@@ -10,14 +10,24 @@ type Employee record {|
...
     int[] salaries = from var {salary} in employees
                      select salary;
 
 
     io:println(salaries);
+
+    // `from` clause iterates `employees` `table` and creates a `table`.
```
MaryamZi, line 26: Shall we rephrase this also? Given that it is the query expression that creates the table rather than the from clause?

examples/querying-tables/querying_tables.bal
```diff
@@ -10,14 +10,24 @@ type Employee record {|
...
+    table<Employee> smiths = from Employee emp in employees
+                     where emp.lastName == "Smith" select emp;
+
+    foreach Employee emp in smiths {
+        io:println(emp.firstName + " " + emp.lastName);
```
MaryamZi, line 31:
```suggestion
        io:println(emp.firstName, " ", emp.lastName);
```

examples/querying-tables/querying_tables.bal
KRVPerera: done

examples/querying-tables/querying_tables.bal
```diff
@@ -10,14 +10,26 @@ type Employee record {|
...
-        {id: 2, firstName: "Fred", lastName: "Bloggs", salary: 200}
+        {id: 2, firstName: "Jane", lastName: "Smith", salary: 150},
+        {id: 4, firstName: "Fred", lastName: "Bloggs", salary: 200},
+        {id: 7, firstName: "Bobby", lastName: "Clark", salary: 300},
+        {id: 9, firstName: "Cassie", lastName: "Smith", salary: 250}
     ];
 
-    // `from` clause iterates `employees` `table`.
-    // The contextually-expected type of the `query expression` is an `int[]`.
+    // The `from` clause iterates `employees` table.
```
pubudu91, line 19:
```suggestion
    // The `from` clause iterates the `employees` table.
```

examples/querying-tables/querying_tables.bal
```diff
@@ -10,14 +10,26 @@ type Employee record {|
...
+    // The query expression creates an integer `array` based on the contextually-expected type.
```
pubudu91, line 20: This is slightly misleading. Gives the impression that it always gives an int array based on the contextually expected type. This should be phrased in a way that conveys what happens in this particular context because I don't think we can fit in the complete behaviour of a query expression when it comes to determining the type of the query expression. So in this case, there's a contextually expected type for the query expr, which is `int[]`, and therefore, it will create an `int[]` value upon its evaluation.

pubudu91: Maybe something like "Since an integer array is contextually expected, the evaluation of the query expression will result in an integer array"?

pubudu91: @MaryamZi thoughts?

MaryamZi: How about
```suggestion
    // Since the contextually-expected type for the query expression is `int[]` evaluation of the query expression will result in an integer array.
```

pubudu91 (approved):

PR #2559, opened 2022-01-29
Fix few bbe descriptions and formatting

lochana-chathura:
## Purpose
$subject.

### DIFF
--- a/examples/check-semantics/check_semantics.bal
+++ b/examples/check-semantics/check_semantics.bal
@@ -3,7 +3,7 @@ import ballerina/io;
 public function main() returns error? {
     do {
         // If either `foo()` or `bar()` invocations returns an `error`,
-        // the error will be returned from the `main` function and execution
+        // the `error` will be returned from the `main` function and execution
         // of the `main` function ends.
         check foo();
         check bar();
@@ -14,7 +14,7 @@ public function main() returns error? {
 
         }
     }
-    // Failure with the respective error is caught by the `on fail` block.
+    // Failure with the respective `error` is caught by the `on fail` block.
     on fail var e {
         io:println(e.toString());
         return e;
--- a/examples/check-semantics/check_semantics.bal
+++ b/examples/check-semantics/check_semantics.bal
@@ -2,19 +2,19 @@ import ballerina/io;
 
 public function main() returns error? {
     do {
-        // If either `foo()` or `bar()` invocations returns an `error`,
-        // the `error` will be returned from the `main` function and execution
+        // If either `foo()` or `bar()` invocations returns an error,
+        // the error will be returned from the `main` function and execution
         // of the `main` function ends.
         check foo();
         check bar();
 
         if !isOK() {
-            // Fails explicitly with an `error`.
+            // Fails explicitly with an error.
             fail error("not OK");
 
         }
     }
-    // Failure with the respective `error` is caught by the `on fail` block.
+    // Failure with the respective error is caught by the `on fail` block.
     on fail var e {
         io:println(e.toString());
         return e;
--- a/examples/check-semantics/check_semantics.description
+++ b/examples/check-semantics/check_semantics.description
@@ -1,4 +1,4 @@
-// `check` semantics is not simply to return on an `error` value.
+// `check` semantics is not to simply return on an `error` value.
 // When `check` gets an `error` value, it fails.
 // The enclosing block decides how to handle the failure.
 // Most blocks pass the failure up to the enclosing block.
--- a/examples/check-semantics/check_semantics.description
+++ b/examples/check-semantics/check_semantics.description
@@ -1,4 +1,4 @@
-// `check` semantics is simply not to return on an error.
+// `check` semantics is not simply to return on an `error` value.
 // When `check` gets an `error` value, it fails.
 // The enclosing block decides how to handle the failure.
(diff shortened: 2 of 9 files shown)


examples/transaction-statement/transaction_statement.description
```diff
@@ -1,5 +1,5 @@
 // Ballerina provides support for interacting with a transaction manager.
-// Compile-time guarantees that transactions are bracketed with begin and commit or rollback.
-// The region in the middle, is typed as being a transactional context.
-// Ballerina does not have a transactional memory and includes transaction manager.
-// Current transaction is a part of the execution context of a strand.
+// Compile-time guarantees that transactions are bracketed with `begin` and `commit` or `rollback`.
```
lochana-chathura, line 2: Not sure about `begin` @pcnfernando @gimantha can you verify?

praneesha (approved):

examples/check-semantics/check_semantics.bal
```diff
@@ -3,7 +3,7 @@ import ballerina/io;
 public function main() returns error? {
     do {
         // If either `foo()` or `bar()` invocations returns an `error`,
-        // the error will be returned from the `main` function and execution
+        // the `error` will be returned from the `main` function and execution
```
MaryamZi, line 6: We emphasize keywords or identifiers. So if we are emphasizing error here we'll have to say something like
```suggestion
        // the `error` value will be returned from the `main` function and execution
```

examples/check-semantics/check_semantics.description
```diff
@@ -1,8 +1,8 @@
-// `check` semantics is not simply to return on error.
-// When `check` gets an error, it fails.
-// Enclosing block decide how to handle failure.
-// Most blocks pass failure up to enclosing block.
-// Function definition handles failure by returning the error.
-// `on fail` can catch the error.
-// `fail` statement is like `check` but always fails.
-// Differs from exceptions in that control flow is explicit
+// `check` semantics is simply not to return on an `error`.
```
MaryamZi, line 1: Wasn't not in the correct place?

Same comment re: emphasizing error.

examples/check-semantics/check_semantics.description
```diff
@@ -1,8 +1,8 @@
...
+// When `check` gets an `error`, it fails.
+// Enclosing block decides how to handle the failure.
```
MaryamZi, line 3:
```suggestion
// The enclosing block decides how to handle the failure.
```

examples/rollback/rollback.bal
```diff
@@ -8,13 +8,15 @@ public function main() returns error? {
...
+    [
+        {updateIndex: 0, stockMnt: 2000},
+        {updateIndex: 1, stockMnt: -1000},
+        {updateIndex: 2, stockMnt: 1500},
+        {updateIndex: 3, stockMnt: -1000},
+        {updateIndex: 4, stockMnt: -2000}
+    ];
+    // If an `error` is returned from the `transfer` function,
+    // the `error` is returned from the `main` and it exits.
     check transfer(updates);
```
MaryamZi, line 20:
```suggestion
    check transfer(updates);
    
```

examples/transaction-statement/transaction_statement.description
MaryamZi: There's no begin keyword, right? We shouldn't be emphasizing begin.

examples/check-semantics/check_semantics.description
lochana-chathura: didn't get what you mean by  `Wasn't in the correct place?`. Please check with the new changes.

MaryamZi: Sorry, meant to say the previous version was correct.

Maybe
```
// `check` semantics is not simply to return on an `error` value.
```

lochana-chathura: reading `not simply to return` felt a bit odd to me. Anyway, updated according to the above suggestion. thanks.

MaryamZi: Or actually maybe
```
// `check` semantics is not to simply return on an `error` value.
```

@praneesha, wdyt? What we want to say is "`check` doesn't just mean returning when the expression evaluates to error".

lochana-chathura: +1 on the new suggestion. Let's see if praneesha has anything to add.

examples/check-semantics/check_semantics.description
```diff
@@ -1,8 +1,8 @@
-// `check` semantics is not simply to return on error.
-// When `check` gets an error, it fails.
-// Enclosing block decide how to handle failure.
-// Most blocks pass failure up to enclosing block.
-// Function definition handles failure by returning the error.
+// `check` semantics is not simply to return on an `error` value.
```
lochana-chathura, line 1:
```suggestion
// `check` semantics is not to simply return on an `error` value.
```

pubudu91:
Are we good to merge this?

MaryamZi (approved):

examples/check-semantics/check_semantics.description
lochana-chathura: Resolving the conversation as the new suggestion was added.

lochana-chathura:
> Are we good to merge this?
yes!.

PR #2566, opened 2022-01-29
Revert spec deviations

KRVPerera:
## Purpose
Some changes are invalid according to spec this PR reverts them

### DIFF
--- a/examples/creating-tables-with-query/creating_tables_with_query.bal
+++ b/examples/creating-tables-with-query/creating_tables_with_query.bal
@@ -14,7 +14,7 @@ public function main() {
     ];
 
     // The query expression starts with `table`.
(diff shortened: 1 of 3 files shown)


examples/creating-tables-with-query/creating_tables_with_query.bal
```diff
@@ -14,7 +14,7 @@ public function main() {
     ];
 
     // The query expression starts with `table`.
-    // The key specifier `key(id)` specifies the key constraint of the constructed `table`.
+    // The key specifier `key(id)` specifies the key sequence of the constructed `table`.
     // The result of the query expression is a `table`.
```
KRVPerera, line 18: > If the query-expr starts with table, then the query-expr will construct a table; the key-specifier specifies the key sequence of the constructed

pubudu91 (approved):

PR #2568, opened 2022-01-30
Revert pr 2550

KavinduZoysa:
## Purpose
$subject


MaryamZi (approved):
Let's create an issue for this also. If we can fix for GA, that would be great.

PR #2577, opened 2022-02-02
Remove exit code for devcontainer

keizer619:


NipunaRanasinghe (commented):
Is there any particular reason to set the exit code as 1 in `L156`? or should it also be removed?

hevayo (approved):

Issue #2484, opened 2022-01-19
Run update tool tests on installers


NipunaMadhushan:
Currently installer tests are running on installers but update tool tests needs to be run on installers in different operating systems.

PR #2585, opened 2022-02-10
Enable update tool tests

NipunaMadhushan:
## Purpose
Enable update tool tests on different operating systems (Ubuntu, MacOS, Windows).

## Goals
Increase the test coverage.

## Issue
https://github.com/ballerina-platform/ballerina-distribution/issues/2484

## Note
Update tool tests can be run in centos as well. But it cannot be automated since there is no runner in github actions for centos and, we can only use a centos container on ubuntu runner. Users are facing the following issue when they use the centos container in github actions.
https://github.com/ballerina-platform/ballerina-update-tool/issues/218


NipunaMadhushan:
Closing this PR since the changes have been added with a another PR https://github.com/ballerina-platform/ballerina-distribution/pull/3058

PR #2807, opened 2022-05-25
Pack minor version dependencies of standard libraries

udda1996:
## Purpose
To pack transitive minor version dependencies of Standard Libraries along with the distribution.  

## Goals
fixes ballerina-platform/ballerina-lang#36047

### DIFF
--- a/ballerina/build.gradle
+++ b/ballerina/build.gradle
@@ -1209,10 +1209,30 @@ task generateCache(dependsOn: ':cache-generator:build') {
 //            If so, pull it from central and store in the extracted-distributions/module directory
 //            generate caches for those modules as well
         }
-        println(dependencyMap)
+        dependencyMap.each { key, val ->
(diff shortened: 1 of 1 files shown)


ballerina/build.gradle
```diff
@@ -1174,6 +1186,59 @@ task generateCache(dependsOn: ':cache-generator:build') {
...
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+        }
+        // generate caches for minor versions of Standard Libraries
+        dependencyMap.each { key, val ->
+            if (val.size > 1 && key != "observe-ballerina-zip") {
```
keizer619, line 1209: This will skip modules only has 1 dependency, is it intentional?

udda1996: Yes. The first one is the version available in the pack

keizer619 (approved):

PR #2823, opened 2022-05-30
Fix regex and xmldata BBEs output verification failure

kalaiyarasiganeshalingam:
## Purpose
$Subect

### DIFF
--- a/ballerina/build.gradle
+++ b/ballerina/build.gradle
@@ -878,6 +878,7 @@ task testExamples() {
                 str = it.trim()
                 if (str != "Compiling source" && str != "${balName}.bal" && str != "Running executable" && str != ""
                         && str != "Picked up JAVA_TOOL_OPTIONS: -Dfile.encoding=UTF8" && !str.startsWith('HINT') &&
+                        // Skip logs for downloading the packages from the central
                         !str.contains("ballerina/")) {
                     bbeOutput.add(str)
                 }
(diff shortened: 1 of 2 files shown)


ballerina/build.gradle
```diff
@@ -877,7 +877,8 @@ task testExamples() {
             lines.each {
                 str = it.trim()
                 if (str != "Compiling source" && str != "${balName}.bal" && str != "Running executable" && str != ""
-                        && str != "Picked up JAVA_TOOL_OPTIONS: -Dfile.encoding=UTF8" && !str.startsWith('HINT')) {
+                        && str != "Picked up JAVA_TOOL_OPTIONS: -Dfile.encoding=UTF8" && !str.startsWith('HINT') &&
+                        !str.contains("ballerina/")) {
```
gimantha, line 881: Shall we put a comment saying why we had to add this line?

azinneera: Wouldn't using `pulled from central successfully` to skip be better?

kalaiyarasiganeshalingam: We have to skip the following logs
`ballerina/xmldata:2.2.2 [central.ballerina.io -> home repo]  100% [==========]`
`ballerina/xmldata:2.2.2 pulled from central successfully`
They both have the same value as `ballerina/`. So I used this.

gimantha (approved):
LGTM

PR #2898, opened 2022-06-27
Update md and bal files for mappings bbes

prakanth97:

### DIFF
--- a/examples/maps/maps.bal
+++ b/examples/maps/maps.bal
@@ -15,7 +15,7 @@ public function main() {
     // Adds a new entry for `z`.
     m["z"] = 5;
 
-    // Using `m["x"]` wouldn't work here because type would be `int?`, not `int`.
+    // Using `m["x"]` wouldn't work here because the type would be `int?` and not `int`.
     m["z"] = m.get("x");
 
 }
--- a/examples/maps/maps.bal
+++ b/examples/maps/maps.bal
@@ -15,8 +15,7 @@ public function main() {
     // Adds a new entry for `z`.
     m["z"] = 5;
 
-    // Using `m["x"]` wouldn't work here because type would be `int?`,
-    // not `int`.
+    // Using `m["x"]` wouldn't work here because type would be `int?`, not `int`.
     m["z"] = m.get("x");
 
 }
--- a/examples/maps/maps.md
+++ b/examples/maps/maps.md
@@ -1,10 +1,6 @@
 # Maps
 
-The `map<T>` type is a mapping from strings to `T`. The map syntax is similar to JSON. Maps are mutable.
-`m[k]` is an "lvalue". `foreach` will iterate over the values of the `map`. `m[k]` gets entry for `k`; `nil` if
-missing. Use `m.get(k)` when you know that there is an entry for `k`. `m.keys()` can be used
-to iterate over keys to get the keys as an `array` of strings. `==` and `!=` on maps is deep. Two maps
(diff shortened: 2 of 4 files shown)


examples/maps/maps.bal
```diff
@@ -15,8 +15,7 @@ public function main() {
     // Adds a new entry for `z`.
     m["z"] = 5;
 
-    // Using `m["x"]` wouldn't work here because type would be `int?`,
-    // not `int`.
+    // Using `m["x"]` wouldn't work here because type would be `int?`, not `int`.
```
praneesha, line 18:
```suggestion
    // Using `m["x"]` wouldn't work here because the type would be `int?` and not `int`.
```

examples/maps/maps.md
```diff
@@ -1,10 +1,6 @@
 # Maps
 
-The `map<T>` type is a mapping from strings to `T`. The map syntax is similar to JSON. Maps are mutable.
-`m[k]` is an "lvalue". `foreach` will iterate over the values of the `map`. `m[k]` gets entry for `k`; `nil` if
-missing. Use `m.get(k)` when you know that there is an entry for `k`. `m.keys()` can be used
-to iterate over keys to get the keys as an `array` of strings. `==` and `!=` on maps is deep. Two maps
-are equal if they have the same set of keys and the values for each key are equal.
+The `map<T>` type is a mapping from strings to `T`. The map syntax is similar to JSON. Maps are mutable. `m[k]` is an "lvalue". `foreach` will iterate over the values of the `map`. `m[k]` gets entry for `k`; `nil` if missing. Use `m.get(k)` when you know that there is an entry for `k`. `m.keys()` can be used to iterate over keys to get the keys as an `array` of strings. `==` and `!=` on maps is deep. Two maps are equal if they have the same set of keys and the values for each key are equal.
```
praneesha, line 3:
```suggestion
The `map<T>` type is a mapping from strings to `T`. The map syntax is similar to JSON. Maps are mutable. `m[k]` is an `lvalue`. `foreach` will iterate over the values of the `map`. `m[k]` gets an entry for `k`; if `nil` is missing. Use `m.get(k)` when you know that there is an entry for `k`. `m.keys()` can be used to iterate over keys to get the keys as an `array` of strings. `==` and `!=` on maps is deep. Two maps are equal if they have the same set of keys and the values of each key are equal.
```

examples/records/records.md
```diff
@@ -1,9 +1,6 @@
 # Records
 
-A `record` type has specific named fields. Fields can be accessed with `r.x`. Records are mutable. `r.x` is an
-`lvalue`. Records can be constructed using a similar syntax to a `map`. Typically, a `record` type is combined
-with a type definition. The name of the type is not significant. A `record` is just a collection of fields.
-Record equality works the same as `map` equality.
+A `record` type has specific named fields. Fields can be accessed with `r.x`. Records are mutable. `r.x` is an `lvalue`. Records can be constructed using a similar syntax to a `map`. Typically, a `record` type is combined with a type definition. The name of the type is not significant. A `record` is just a collection of fields. Record equality works the same as `map` equality.
```
praneesha, line 3:
```suggestion
A `record` type has specific named fields. These fields can be accessed with `r.x`. Records are mutable. `r.x` is an `lvalue`. Records can be constructed using a similar syntax to a `map`. Typically, a `record` type is combined with a type definition. The name of the type is not significant. A `record` is just a collection of fields. Record equality works the same as `map` equality.
```

praneesha (approved):

PR #2927, opened 2022-06-28
Update `Work directly with JSON` BBE MD File

praneesha:
## Purpose
Update `Work directly with JSON` BBE MD file.

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks


examples/working-directly-with-json/working_directly_with_json.md
```diff
@@ -1,11 +1,10 @@
-# Working directly with JSON
+# Work directly with JSON
```
MaryamZi, line 1: We have multiple BBEs that have similar titles. E.g., [Ignoring return values and errors](https://ballerina.io/learn/by-example/ignoring-return-values-and-errors.html), [Controlling openness](https://ballerina.io/learn/by-example/controlling-openness.html), etc. Can we leave these as is and do the changes in a separate PR if required?

examples/working-directly-with-json/working_directly_with_json.out
```diff
@@ -1,3 +1,3 @@
-bal run working_directly_with_json.bal
+$ bal run working_directly_with_json.bal
```
MaryamZi, line 1: `$` was removed after some discussions IIRC. Shall we finalize if we are adding it back before changing it here?

praneesha: @anupama-pathirage @sameerajayasoma

examples/working-directly-with-json/working_directly_with_json.md
praneesha: Fixed.

examples/working-directly-with-json/working_directly_with_json.out
MaryamZi: I prefer having it too btw. And as discussed, since with BBEs the user may not generally have the same file name, it's highly unlikely they'll copy the command as is. So copying shouldn't be an issue here IMO. We'll also have the output in the same file which is probably why we don't give the option to copy atm?

anupama-pathirage: What if we add it later, when we figure out things (like having a way to copy without the `$` sign etc)

praneesha: Thanks, Yasith! Will keep the $sign as we will be implementing this fix soon.

anupama-pathirage (approved):

MaryamZi (approved):

examples/working-directly-with-json/working_directly_with_json.out
MaryamZi: I was just thinking maybe @keizer619's comment about reverting (deleting the description file) in a single PR would apply here too? Should we do this change for all BBEs also in a single PR?

praneesha: We have already merged the fix of this and now the $sign isn't getting copied. Therefore, we can finalize adding this.

PR #2983, opened 2022-07-18
Remove the `-ing` form from all BBE titles

praneesha:
## Purpose
Remove the `-ing` form from all BBE titles.

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

### DIFF
--- a/examples/index.json
+++ b/examples/index.json
@@ -6,7 +6,7 @@
         "samples": [
             {
                 "name": "Hello world main",
-                "url": "hello-world-main",
+                "url": "hello-world",
                 "verifyBuild": true,
                 "verifyOutput": true,
                 "isLearnByExample": true
@@ -211,7 +211,7 @@
                 "isLearnByExample": true
             },
             {
-                "name": "Error reporting",
+                "name": "Errors",
                 "url": "error-reporting",
                 "verifyBuild": true,
                 "verifyOutput": true,
@@ -550,8 +550,8 @@
                 "isLearnByExample": true
             },
             {
-                "name": "tables querying",
-                "url": "tables-querying",
+                "name": "Querying tables",
+                "url": "querying-tables",
                 "verifyBuild": true,
                 "verifyOutput": true,
                 "isLearnByExample": true
@@ -989,7 +989,7 @@
         "category": "Language concepts",
         "samples": [
             {
-                "name": "assertions",
+                "name": "Assertions",
                 "url": "testerina-assertions",
                 "verifyBuild": false,
                 "verifyOutput": false,
--- a/examples/index.json
+++ b/examples/index.json
@@ -536,7 +536,7 @@
                 "isLearnByExample": true
             },
             {
-                "name": "Order",
+                "name": "Ordering",
                 "url": "ordering",
                 "verifyBuild": true,
                 "verifyOutput": true,
--- a/examples/index.json
+++ b/examples/index.json
@@ -701,7 +701,7 @@
         "samples": [
             {
                 "name": "Consume services: client objects",
-                "url": "consuming-services-client-objects",
+                "url": "consuming-services",
                 "verifyBuild": true,
                 "verifyOutput": false,
                 "disableVerificationReason": "Failing intermittently due to github rate limiting",
--- a/examples/index.json
+++ b/examples/index.json
@@ -295,7 +295,7 @@
                 "isLearnByExample": true
             },
             {
-                "name": "Controlling openness",
+                "name": "Control openness",
                 "url": "controlling-openness",
                 "verifyBuild": true,
                 "verifyOutput": true,
@@ -316,7 +316,7 @@
                 "isLearnByExample": true
             },
             {
-                "name": "Defining classes",
+                "name": "Define classes",
                 "url": "defining-classes",
                 "verifyBuild": true,
                 "verifyOutput": true,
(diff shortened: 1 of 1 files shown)


gimantha (approved):

examples/index.json
```diff
@@ -295,7 +295,7 @@
                 "isLearnByExample": true
             },
             {
-                "name": "Controlling openness",
+                "name": "Control openness",
                 "url": "controlling-openness",
```
MaryamZi, line 299: Shouldn't these also change?

examples/index.json
```diff
@@ -536,7 +536,7 @@
                 "isLearnByExample": true
             },
             {
-                "name": "Ordering",
+                "name": "Order",
```
MaryamZi, line 539: I don't think this makes sense. "Ordering" here is used as a noun.

MaryamZi (commented):
@praneesha, why do we need this change? In some places it feels natural to have -ing.

praneesha:
> @praneesha, why do we need this change? In some places it feels natural to have -ing.

@MaryamZi - We have removed the `ing` form in all the titles in all components of Ballerina documentation and this is to be consistent with that effort. This is done as we are complying with the Google style guide for Ballerina docs: https://developers.google.com/style/headings?hl=en

examples/index.json
praneesha: Yes. `Controlling openness` has been changed to `Control openness`. Do you mean we should change the URL as well?

examples/index.json
praneesha: Reverted to `Ordering`.

examples/index.json
MaryamZi: Yes.

praneesha: Updated the URLs.

examples/index.json
```diff
@@ -6,7 +6,7 @@
         "samples": [
             {
                 "name": "Hello world main",
-                "url": "hello-world",
+                "url": "hello-world-main",
```
MaryamZi, line 9: This isn't necessary, right?

examples/index.json
```diff
@@ -211,7 +211,7 @@
                 "isLearnByExample": true
             },
             {
-                "name": "Errors",
+                "name": "Error reporting",
```
MaryamZi, line 214: Why are we changing these? I wouldn't change the titles to these.

examples/index.json
```diff
@@ -550,15 +550,15 @@
                 "isLearnByExample": true
             },
             {
-                "name": "Querying tables",
-                "url": "querying-tables",
+                "name": "tables querying",
```
MaryamZi, line 553: Why are we changing this?

MaryamZi: This is a title btw.

examples/index.json
```diff
@@ -989,7 +989,7 @@
         "category": "Language concepts",
         "samples": [
             {
-                "name": "Assertions",
+                "name": "assertions",
```
MaryamZi, line 992: Why are we starting with a lower case letter?

examples/index.json
praneesha: Reverted the URL to `hello-world`.

examples/index.json
praneesha: Changed to be consistent with the URL. Reverted the name back to `Errors`.

examples/index.json
praneesha: Reverted the name tback o `Querying tables` and the URL to `querying-tables`.

examples/index.json
praneesha: Sorry, that was a mistake. Fixed.

examples/index.json
```diff
@@ -444,22 +444,22 @@
...
+                "url": "convert-from-user-defined-type-to-json",
                 "verifyBuild": true,
                 "verifyOutput": true,
                 "isLearnByExample": true
             },
             {
-                "name": "Converting from JSON to user-defined type",
-                "url": "converting-from-json-to-user-defined-type",
+                "name": "Convert from JSON to user-defined type",
+                "url": "convert-from-json-to-user-defined-type",
```
MaryamZi, line 455: These will result in quite a few broken links, right? I'm not sure if that's acceptable given the number of BBEs we're changing.

examples/index.json
```diff
@@ -700,15 +700,15 @@
         "category": "Language concepts",
         "samples": [
             {
-                "name": "Consuming services: client objects",
-                "url": "consuming-services",
+                "name": "Consume services: client objects",
+                "url": "consuming-services-client-objects",
```
MaryamZi, line 704: Not required?

examples/index.json
praneesha: Reverted the URL changes.

examples/index.json
praneesha: Fixed.

examples/index.json
```diff
@@ -578,7 +578,7 @@
                 "isLearnByExample": true
             },
             {
-                "name": "Querying with streams",
+                "name": "Query with streams",
```
MaryamZi, line 581: Here again I think "Querying" is more suitable.

MaryamZi (commented):
Lang changes LGTM.

examples/index.json
praneesha: Reverted.

shafreenAnfar (approved):
I don't see any technical issues here.

examples/index.json
```diff
@@ -975,7 +975,7 @@
                 "isLearnByExample": true
             },
             {
-                "name": "Inferring isolated",
+                "name": "Infer isolated",
```
MaryamZi, line 978: For this and L950 also, I think what we had previously is better.

praneesha: Reverted both.

MaryamZi (approved):
Lang BBE changes LGTM.

PR #3013, opened 2022-07-26
Add new xmldata BBE to the conversion between XML and record

kalaiyarasiganeshalingam:
## Purpose
$Subject

### DIFF
--- a/examples/xml-from-record-conversion/xml_from_record_conversion.out
+++ b/examples/xml-from-record-conversion/xml_from_record_conversion.out
@@ -1,0 +1,2 @@
+bal run xml_from_record_conversion.bal
+<ns:Invoice xmlns="example.com" xmlns:ns="http://sdf.com" status="paid"><id>1</id><items xmlns="http://example1.com"><itemCode>223345</itemCode><count>1</count></items><items xmlns="http://example1.com"><itemCode>223300</itemCode><count>7</count></items></ns:Invoice>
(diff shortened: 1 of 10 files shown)


examples/xml-from-record-conversion/xml_from_record_conversion.out
```diff
@@ -0,0 +1,2 @@
+bal run xml_from_record_conversion.bal
```
praneesha, line 1:
```suggestion
$ bal run xml_from_record_conversion.bal
```

Please check and add the $ sign in all the commands added in all `.out` files.

examples/xml-to-record-conversion/xml_to_record_conversion.out
```diff
@@ -0,0 +1,2 @@
+bal run xml_to_record_conversion.bal
```
praneesha, line 1:
```suggestion
bal run xml_to_record_conversion.bal
```
Same as above.

examples/xml-from-record-conversion/xml_from_record_conversion.out
kalaiyarasiganeshalingam: Do we need to add the $ sign to the master repo BBE?

praneesha: Sorry, me bad. We need not.

examples/xml-to-record-conversion/xml_to_record_conversion.out
praneesha: Sorry, we need not add the $sign on the master branch as it is available only on the new website.

praneesha (approved):

PR #3084, opened 2022-08-18
Add few new lang BBEs

MaryamZi:
## Purpose
Add the following BBEs.
- Built-in integer subtypes
- Built-in string subtype
- Computed field key
- Type intersection for error types


examples/built-in-integer-subtypes/built_in_integer_subtypes.out
```diff
@@ -0,0 +1,6 @@
+bal run built_in_integer_subtypes.bal
```
SasinduDilshara, line 1: This should be `$ bal run` ryt?

examples/built-in-integer-subtypes/built_in_integer_subtypes.out
```diff
@@ -0,0 +1,6 @@
...
+72
+43543
+error: {ballerina}NumberConversionError {"message":"'int' value '4,294,967,296' cannot be converted to 'lang.int:Unsigned32'"}
```
SasinduDilshara, line 4: If user run this code, he will get a runtime error. Is it ok?
Isn't it better if we catch the error in the Ballerina Code and print `'int' value '4,294,967,296' cannot be converted to 'lang.int:Unsigned32'`.

examples/built-in-string-subtype/built_in_string_subtype.bal
```diff
@@ -0,0 +1,13 @@
+import ballerina/io;
+
+public function main() {
+    // A variable of type `string:Char` to which only strings of
```
KRVPerera, line 4: Can we keep the comment in one line? 91 Columns as of now

examples/built-in-string-subtype/built_in_string_subtype.md
```diff
@@ -0,0 +1,11 @@
+# Built-in string subtype
+
+The `ballerina/lang.string` lang library defines the `Char` type, which is a built-in subtype of `string`. A string belongs to the `string:Char` type if it has length 1.
```
KRVPerera, line 3: Cant we add a link here?

examples/computed-field-key/computed_field_key.md
```diff
@@ -0,0 +1,8 @@
+# Computed field key
+
+Ballerina allows you to have computed values as keys in a map. This can be done by specifying the key as an expression within square brackets.
+This is particularly useful when you want to define and use constants for key values.
```
KRVPerera, line 4: I think this second sentence reduce the value of the feature or it does not convey your idea yet.

examples/error-type-intersection/error_type_intersection.bal
```diff
@@ -0,0 +1,33 @@
+import ballerina/io;
+
+type IoError distinct error;
```
KRVPerera, line 3: May need change in many places
```suggestion
type IOError distinct error;
```

examples/error-type-intersection/error_type_intersection.bal
```diff
@@ -0,0 +1,33 @@
...
+    // `fileIoError` belongs to both `IoError` and `error<FileErrorDetail>`.
+    io:println(fileIoError is IoError);
+    io:println(fileIoError is error<FileErrorDetail>);
+
+    // An `IoError` value will not belong to `FileIoError` if it doesn't belong to
+    // `error<FileErrorDetail>`.
+    IoError ioError = error("invalid input");
+    io:println(ioError is FileIoError);
+
+    // Similarly an error value belonging to `error<FileErrorDetail>` will not belong
```
KRVPerera, line 29:
```suggestion
    // Similarly, an error value belonging to `error<FileErrorDetail>` will not belong to `FileIoError` if it doesn't also belong to `IoError`.
```

examples/error-type-intersection/error_type_intersection.md
```diff
@@ -0,0 +1,7 @@
+# Type intersection for error types
+
+You can define an error type that is both a subtype of a `distinct` error type and has additional constraints on the detail fields using intersection types.
```
KRVPerera, line 3:
```suggestion
You can define an error type that is a subtype of a `distinct` error type and has additional constraints on the detail fields using intersection types.
```

examples/built-in-integer-subtypes/built_in_integer_subtypes.out
MaryamZi: It's okay IMO. Panics are part of the language. Trapping it will bring in extra (unrelated) concepts to the BBE.

examples/built-in-string-subtype/built_in_string_subtype.md
MaryamZi: Does it add any value though? https://lib.ballerina.io/ballerina/lang.string/0.0.0/types#Char is what we have.

examples/computed-field-key/computed_field_key.md
MaryamZi: Any particular reason? 

This is from the [slides](https://ballerina.io/learn/slides/language-walkthrough/Ballerina_Language_Presentation-2021-03-08.pdf). This is a commonly asked question though. People usually use constants directly as keys and wonder why it doesn't work in the way they expect.

examples/error-type-intersection/error_type_intersection.md
MaryamZi: Any particular reason to remove "both"?

examples/error-type-intersection/error_type_intersection.bal
MaryamZi: Added the comma, but merging the lines exceeds the column count.

examples/error-type-intersection/error_type_intersection.md
KRVPerera: One is a type, and the other is a constraint on the type. Does both match here? Fine to keep it too

examples/computed-field-key/computed_field_key.md
KRVPerera: okay. I think we give too much emphasis to that here that is all.  What if we remove the word `particularly`?

examples/built-in-string-subtype/built_in_string_subtype.md
KRVPerera: To the string module. https://lib.ballerina.io/ballerina/lang.string/0.0.0 . Where we mention it in the first sentence. We may have to consistent across BBEs

MaryamZi: This was considered previously also, but we didn't go ahead with it since it wasn't possible for most BBEs. Unlike with stdlib BBEs lang BBEs don't always correspond to a langlib or they may have multiple relevant langlibs. There's also the possibility of users not knowing the concept of langlibs when they get to certain BBEs.

Anyway since this BBE is about `string:Char` don't think adding a link to the string langlib here is the best thing. Would be more relevant if we did it for the string BBE, but that's one of the first BBEs and it's possible that the user does not know of langlibs when they get there, if they are following the BBEs in order.

examples/computed-field-key/computed_field_key.md
MaryamZi: Does it make a difference in the emphasis though? How about 

```suggestion
This feature can be used when you want to define and use constants for key values.
```

examples/error-type-intersection/error_type_intersection.md
MaryamZi: What we are trying to say is 

```
You can define an error type
- that is a subtype of a `distinct` error type
- that has additional constraints on the detail fields
```

So "both" here is OK IMO.

examples/computed-field-key/computed_field_key.md
KRVPerera: Yes, this is good. Maybe this is just a subjective comment. I am okay to keep it as it is or for the new one.

examples/built-in-integer-subtypes/built_in_integer_subtypes.md
```diff
@@ -0,0 +1,20 @@
+# Built-in integer subtypes
+
+Ballerina has one integer type (`int`) which is a 64-bit, signed type.
```
praneesha, line 3:
```suggestion
Ballerina has one integer type (`int`), which is a 64-bit, signed type. 
```

examples/built-in-integer-subtypes/built_in_integer_subtypes.md
```diff
@@ -0,0 +1,20 @@
...
+The `ballerina/lang.int` lang library defines the following built-in subtypes of this `int` type.
+
+- `int:Signed32`
+- `int:Unsigned32`
+- `int:Signed16`
+- `int:Unsigned16`
+- `int:Signed8`
+- `int:Unsigned8` (same as `byte`)
+
+Runtime behaviour of operations on these subtypes are the same as for the `int` type. Bitwise operations with operands of these types have special typing which allow using a more specific type for the operation.
```
praneesha, line 14:
```suggestion
The runtime behavior of operations on these subtypes are the same as for the `int` type. Bitwise operations with operands of these types have special typing, which allows using a more specific type for the operation.
```

examples/built-in-string-subtype/built_in_string_subtype.md
praneesha, line 3:
```suggestion
The `ballerina/lang.string` lang library defines the `Char` type, which is a built-in subtype of `string`. A string belongs to the `string:Char` type if it has a length of 1.
```

examples/computed-field-key/computed_field_key.bal
```diff
@@ -0,0 +1,23 @@
...
+
+const ID = "unique-identifier";
+
+function getDetailMap(string uniqueId,
+                      string firstName,
+                      string lastName,
+                      string idName, 
+                      string id) returns map<string> {
+    return {
+        // The value `"unique-identifier"` is subtituted from the constant as the key.
```
praneesha, line 11:
```suggestion
        // The `unique-identifier`value is substituted from the constant as the key.
```

examples/error-type-intersection/error_type_intersection.bal
```diff
@@ -0,0 +1,33 @@
...
+};
+
+// The `FileIOError` type is defined as an intersection type using the `&` notation.
+// It is the intersection of two error types: `IOError` and `error<FileErrorDetail>`.
+// An error value belongs to this type if and only if it belongs to both `IOError` 
+// and `error<FileErrorDetail>`.
+type FileIOError IOError & error<FileErrorDetail>;
+
+public function main() {
+    // In order to create an error value that belongs to `FileIOError` the `filename`
```
praneesha, line 16:
```suggestion
    // In order to create an error value that belongs to `FileIOError`, the `filename`
```

examples/error-type-intersection/error_type_intersection.bal
```diff
@@ -0,0 +1,33 @@
...
+    io:println(fileIOError is IOError);
+    io:println(fileIOError is error<FileErrorDetail>);
+
+    // An `IOError` value will not belong to `FileIOError` if it doesn't belong to
+    // `error<FileErrorDetail>`.
+    IOError ioError = error("invalid input");
+    io:println(ioError is FileIOError);
+
+    // Similarly, an error value belonging to `error<FileErrorDetail>` will not belong 
+    // to `FileIOError` if it doesn't also belong to `IOError`.
```
praneesha, line 30:
```suggestion
    // to `FileIOError` if it doesn't belong to `IOError`.
```

examples/computed-field-key/computed_field_key.bal
MaryamZi: Will include quotes to reflect the value.

```suggestion
        // The `"unique-identifier"` value is substituted from the constant as the key.
```

examples/computed-field-key/computed_field_key.md
MaryamZi: Will keep as is then. Thanks!

examples/built-in-integer-subtypes/built_in_integer_subtypes.metatags
```diff
@@ -0,0 +1,2 @@
+description: This BBE demonstrates built-in integer subtypes in Ballerina.
+keywords: ballerina, ballerina by example, bbe, int, int subtypes, signed, unsigned
```
SasinduDilshara, line 2: Shall we add specific signed and unsigned types in here. Because this files includes keywords for searching purposes
ex:- signed32, unsigned32, signed 16, unsigned16.

examples/error-type-intersection/error_type_intersection.metatags
```diff
@@ -0,0 +1,2 @@
+description: This BBE demonstrates type intersection for error types.
+keywords: ballerina, ballerina by example, bbe, error, intersection, distinct
```
SasinduDilshara, line 2: Do we need `distinct` keyword in here

examples/built-in-integer-subtypes/built_in_integer_subtypes.metatags
MaryamZi: @praneesha, is there a limit on the number of keywords we can include?

examples/error-type-intersection/error_type_intersection.metatags
MaryamZi: It's OK to have IMO. This is pretty much the only way to define subtypes of a distinct error type with additional constraints.

praneesha (approved):

examples/built-in-integer-subtypes/built_in_integer_subtypes.metatags
praneesha: @MaryamZi - We don't have a limit but let's not make it a huge list too.

SasinduDilshara: Then, I think the current keywords will enough.

SasinduDilshara (approved):

PR #3226, opened 2022-09-25
Throw Exception when generating API docs from bala

### LINKED ISSUE #3218: Create a gradle task to check whether API docs are generated
**Description:**
Currently API docs are generated and not checked whether they are generated or not. With this check, it will check this scenario.

**Describe your problem(s)**

**Describe your solution(s)**

**Related Issues (optional):**

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

NipunaMadhushan:
## Purpose
Currently if API docs generation fails for a module, it won't be checked when the build goes. This will throw exception instead of giving a log message. Ballerinai modules have been skipped since API docs are not generated for those modules.

## Fixes #3218


dist-repo-builder/src/main/java/io/ballerina/dist/DistRepoBuilder.java
```diff
@@ -63,9 +63,11 @@ public static void main(String args[]) throws IOException {
...
         List<Path> existingDocs = getExistingDocs(jBalToolsPath.resolve("docs"));
         for (Path bala : balas) {
-            generateDocsFromBala(bala, jBalToolsPath, existingDocs);
-            // following function was put in to validate if bir and jar exists for packed balas
-            valid = valid & validateCache(bala, repo);
+            if (!bala.toString().contains("ballerinai")) {
+                generateDocsFromBala(bala, jBalToolsPath, existingDocs);
+                // following function was put in to validate if bir and jar exists for packed balas
+                valid = valid & validateCache(bala, repo);
+            }
```
NipunaRanasinghe, line 70: Shall we use a continue statement below, so that we can clearly separate out the omitting logic?

```suggestion
            if (bala.toString().contains("ballerinai")) {
               continue;
            }
            generateDocsFromBala(bala, jBalToolsPath, existingDocs);
            // following function was put in to validate if bir and jar exists for packed balas
            valid = valid & validateCache(bala, repo);
```

dist-repo-builder/src/main/java/io/ballerina/dist/DistRepoBuilder.java
```diff
@@ -63,9 +63,11 @@ public static void main(String args[]) throws IOException {
         // The following list will contain existing docs from ballerina-lang repo
         List<Path> existingDocs = getExistingDocs(jBalToolsPath.resolve("docs"));
         for (Path bala : balas) {
-            generateDocsFromBala(bala, jBalToolsPath, existingDocs);
-            // following function was put in to validate if bir and jar exists for packed balas
-            valid = valid & validateCache(bala, repo);
+            if (!bala.toString().contains("ballerinai")) {
```
NipunaRanasinghe, line 66: Shall we use a constant here? (i.e: `ORG_BALLERINA_I`)

NipunaRanasinghe, line 66: Let's also add a comment explaining WHY we skip this particular bala org

dist-repo-builder/src/main/java/io/ballerina/dist/DistRepoBuilder.java
NipunaMadhushan: @NipunaRanasinghe addressed the suggestions.

dist-repo-builder/src/main/java/io/ballerina/dist/DistRepoBuilder.java
NipunaMadhushan: @NipunaRanasinghe addressed the suggestions.

NipunaMadhushan: @NipunaRanasinghe addressed the suggestions.

dist-repo-builder/src/main/java/io/ballerina/dist/DistRepoBuilder.java
```diff
@@ -63,6 +64,10 @@ public static void main(String args[]) throws IOException {
         // The following list will contain existing docs from ballerina-lang repo
         List<Path> existingDocs = getExistingDocs(jBalToolsPath.resolve("docs"));
         for (Path bala : balas) {
+            // skipping ballerinai modules since API docs are not generated for those modules
```
NipunaRanasinghe, line 67: Let's update the command along with the new changes

PR #3258, opened 2022-09-29
Skip API docs gen in windows

NipunaMadhushan:
## Purpose
With recent change done with https://github.com/ballerina-platform/ballerina-distribution/pull/3226 will throw exception in windows build since bala paths have more characters than the maximum. This will skip the API docs generation in windows build.


malinthar (approved):
LGTM!

PR #3417, opened 2022-11-11
Improve transaction BBE 

niveathika:
## Purpose
$subject

Fixes: https://github.com/ballerina-platform/ballerina-standard-library/issues/3475


examples/mysql-atomic-transaction/mysql_atomic_transaction.bal
```diff
@@ -13,23 +13,14 @@ public function main() returns error? {
...
-        } else {
-            error? err = commit;
-            if err is error {
-                log:printError("Error occurred while committing", err);
-            }
-        }
+        check commit;
+    } on fail error e {
+        log:printError(e.message());
+        log:printInfo("One of the queries failed. Rollback transaction.");
```
daneshk, line 23: Don't we need to rollback here

examples/jdbc-atomic-transaction/jdbc_atomic_transaction.bal
```diff
@@ -14,23 +13,14 @@ public function main() returns error? {
...
-            }
-        }
+
+        _ = check jdbcClient->execute(`INSERT INTO Customers (firstName, lastName, registrationID, creditLimit, country)
+                                        VALUES ('Peter', 'Stuart', 4, 5000.75, 'USA')`);
+
+        check commit;
+    } on fail error e {
+        log:printError(e.message());
+        log:printInfo("One of the queries failed. Rollback transaction.");
```
daneshk, line 23: Don't we need to rollback here

examples/mysql-atomic-transaction/mysql_atomic_transaction.bal
niveathika: Rollback is called automatically if the statements return error from check statements

examples/jdbc-atomic-transaction/jdbc_atomic_transaction.bal
niveathika: Rollback is called automatically if the statements return error from check statements

daneshk (approved):

Issue #3635, opened 2022-11-29
[BBE Hackathon] Improve SQL BBEs


niveathika:
**Description:**
- [x] Combine JDBC & MySQL BBEs under "Database Access" section
- [x] Convert util functions to standalone bal file that can be referred from the prerequisite section
- [x] Add a tip to point to all other connectors and drivers we support
- [x] Convert simple scenarios from main function to http service
- [x] Change user inputs as either command line inputs or resource inputs
- [x] Converge all BBE use case into one, "Music Store"
- [ ] [Have a BBE explaining the client and connection lifecycle
](https://github.com/ballerina-platform/ballerina-standard-library/issues/3807)

niveathika:
Closing this issue as all the tasks are complete. The remaining task will be tracked seperately.

PR #3696, opened 2022-12-05
Convert Database Access - Simple Query BBE to http service

niveathika:
## Purpose
$subject

Part of https://github.com/ballerina-platform/ballerina-distribution/issues/3635


examples/mysql-query-operation/mysql_simple_query.bal
```diff
@@ -0,0 +1,32 @@
...
+    string title;
+    string artist;
+    float price;
+|};
+
+service / on new http:Listener(8080) {
+    private final mysql:Client db;
+
+    function init() returns error? {
+        // Iniitate the mysql client at the start of the service. This will be used
```
daneshk, line 18:
```suggestion
        // Initiate the mysql client at the start of the service. This will be used
```

PR #3730, opened 2022-12-06
Fix xmldata BBE output verification failure

kalaiyarasiganeshalingam:
## Purpose
$Subject

### DIFF
--- a/ballerina/build.gradle
+++ b/ballerina/build.gradle
@@ -879,7 +879,9 @@ task testExamples() {
             lines.each {
                 str = it.trim()
                 if (str != "Compiling source" && str != "${balName}.bal" && str != "Running executable" && str != ""
-                        && str != "Picked up JAVA_TOOL_OPTIONS: -Dfile.encoding=UTF8" && !str.startsWith('HINT')) {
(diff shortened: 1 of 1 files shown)


HindujaB (approved):

niveathika (commented):
@kalaiyarasiganeshalingam Is this the correct fix? Shouldn't the versions be kept upto date rather than skipping the said logs?

kalaiyarasiganeshalingam:
> Is this the correct fix? Shouldn't the versions be kept upto date rather than skipping the said logs?

Earlier, we faced the same issue in the 2201.0.x and fixed it with this fix.  
https://github.com/ballerina-platform/ballerina-distribution/pull/2823

niveathika:
> Earlier, we faced the same issue in the 2201.0.x and fixed it with this fix.  
> https://github.com/ballerina-platform/ballerina-distribution/pull/2823

Wasnt it a special case? Where we did not want to pull the latest due to a bug in the GA pack?

kalaiyarasiganeshalingam:
> > Earlier, we faced the same issue in the 2201.0.x and fixed it with this fix.
> > #2823
> 
> Wasnt that a special case? Where we did not want to pull the latest due to a bug in the GA pack?

Updated xmldata version in https://github.com/ballerina-platform/ballerina-distribution/pull/3755

PR #3778, opened 2022-12-12
Update HTTP data-binding related BBE descriptions

ayeshLK:
Part of [#3812](https://github.com/ballerina-platform/ballerina-standard-library/issues/3812)


examples/http-service-data-binding/http_service_data_binding.md
```diff
@@ -1,6 +1,6 @@
 # REST service - Payload data binding
 
-HTTP service payload data binding helps to access the request payload through a resource signature parameter. The payload parameter should be declared with the `@Payload` annotation and the  parameter type can be `anydata`. Binding failures will be responded with a 400 Bad Request response.
+HTTP service payload data binding allows accessing the request payload using a resource signature parameter. The resource parameter which is bound to the request payload should be annotated with `@http:Payload` annotation. The resource parameter type should be `anydata`.If the data binding fails, a `400 Bad Request` response will be sent to the client. Use this to access the request payload directly from the resource.
```
chamil321, line 3:
```suggestion
HTTP service payload data binding allows accessing the request payload using a resource signature parameter. The resource parameter which is bound to the request payload should be annotated with `@http:Payload` annotation. The resource parameter type should be `anydata`. If the data binding fails, a `400 Bad Request` response is sent to the client. Use this to access the request payload directly from the resource.
```

examples/http-client-data-binding/http_client_data_binding.md
```diff
@@ -1,6 +1,6 @@
...
+HTTP client payload data binding feature allows accessing the response payload directly without manipulating the `http:Response` object. The response payload type is inferred from the contextually expected type or the `targetType` argument. `anydata` type or `http:Response` is expected as the expected-type or the `targetType` along with `error` type. When the client payload data binding is expected, the HTTP error responses (4xx, 5xx) are categorized as an `error` (`http:ClientRequestError`, `http:RemoteServerError`) of the client's remote operation. If the expected payload type is not compatible with the response payload type an `http:ClientError` will be returned. Use this to access the response payload directly when the response payload type is known beforehand.
```
chamil321, line 3:
```suggestion
HTTP client payload data binding feature allows accessing the response payload directly without manipulating the `http:Response` object. The response payload type is inferred from the contextually expected type or the `targetType` argument. `anydata` type or `http:Response` is expected as the expected-type or the `targetType` along with `error` type. When the client payload data binding is expected, the HTTP error responses (4xx, 5xx) are categorized as an `error` (`http:ClientRequestError`, `http:RemoteServerError`) of the client's remote operation. If the expected payload type is not compatible with the response payload type an `http:ClientError` will be returned. Use this to access the response payload directly when the response payload type is known beforehand. 
```

PR #3788, opened 2022-12-13
Update client data-binding description

ayeshLK:
Part of [#3812](https://github.com/ballerina-platform/ballerina-standard-library/issues/3812)


examples/http-client-data-binding/http_client_data_binding.md
```diff
@@ -1,6 +1,6 @@
...
+The `http:Client` payload data-binding allows directly binding the response payload to a given subtype of `anydata`. It does this by mapping a given HTTP content-type to one or more Ballerina types. For instance, `text/plain` is mapped to `string` whereas `application/json` is mapped to `json`, `record`, etc. Client data-binding can be used by simply assigning the resource method’s returned value to the declared variable. If the response is anything other than 2xx, an `error` is returned and no data-binding is performed. If there is no mapping between the given Ballerina type and the response content-type, again an `error` is returned. Use this when the application is only interested in the response payload but not the headers. When the response payload is JSON, `record` type is preferred to `json` type as it provides compile-time validations, better readability and improved tooling support.
```
TharmiganK, line 3:
```suggestion
The `http:Client` payload data-binding allows directly binding the response payload to a given subtype of `anydata`. It does this by mapping a given HTTP content-type to one or more Ballerina types. For instance, `text/plain` is mapped to `string`, whereas `application/json` is mapped to `json`, `record`, etc. The client data-binding can be used by simply assigning the resource method’s returned value to the declared variable. If the response is anything other than 2xx, an `error` is returned and no data-binding is performed. If there is no mapping between the given Ballerina type and the response content-type, again an `error` is returned. Use this when the application is only interested in the response payload but not the headers. When the response payload is JSON, the `record` type is preferred to the `json` type as it provides compile-time validations, better readability, and improved tooling support.
```

praneesha, line 3:
```suggestion
The `http:Client` payload data binding allows binding the response payload directly to a given subtype of `anydata`. It does this by mapping a given HTTP content-type to one or more Ballerina types. For instance, `text/plain` is mapped to `string` whereas `application/json` is mapped to `json`, `record`, etc. 

Client data binding can be used by simply assigning the resource method’s returned value to the declared variable. If the response is anything other than 2xx, an `error` is returned and no data binding is performed. If there is no mapping between the given Ballerina type and the response content-type, again an `error` is returned. Use this when the application is only interested in the response payload but not the headers. When the response payload is JSON, the `json` type is preferred as the `record` type as it provides compile-time validations, better readability, and improved tooling support.
```
This paragraph is too lengthy. Shall we break it into 2? Also, check if the technical meaning is affected by the suggested change on the `json` type and the `record type`. @ayeshLK

praneesha: Please refer below comment as we need a few updates WRT this suggestion.

shafreenAnfar: It is only 140 words. So one paragraph should be okay, IMO. Remember, when you render it in the website it looks a lot more smaller.

chamil321 (approved):

PR #3891, opened 2022-12-21
Revamp table BBEs

ushirask:

PR #3896, opened 2022-12-21
Change table BBEs

KavinduZoysa:
## Purpose
Fixes #3891


KavinduZoysa:
Addressed in https://github.com/ballerina-platform/ballerina-distribution/pull/3897, hence closing.

PR #3898, opened 2022-12-21
Add GraphQL Service Error Handling BBE

ThisaruGuruge:
## Purpose
$Subject

Fixes: https://github.com/ballerina-platform/ballerina-standard-library/issues/3863

### DIFF
--- a/examples/graphql-client-handle-partial-response/graphql_client_handle_partial_response.bal
+++ b/examples/graphql-client-handle-partial-response/graphql_client_handle_partial_response.bal
@@ -12,7 +12,7 @@ type ProfileResponse record {|
     record {|Profile profile;|} data;
 |};
 
-// The following record type defines the shape of the response from a GraphQL service which allows
+// The following record type defines the shape of the response from a GraphQL service, which allows
 // the `age` field to have a `null` value.
 type Profile record {|
     string name;
--- a/examples/graphql-client-handle-partial-response/graphql_client_handle_partial_response.bal
+++ b/examples/graphql-client-handle-partial-response/graphql_client_handle_partial_response.bal
@@ -13,9 +13,9 @@ type ProfileResponse record {|
 |};
 
 // The following record type defines the shape of the response from a GraphQL service which allows
-// the `name` and `age` fields to have a `null` value.
+// the `age` field to have a `null` value.
 type Profile record {|
-    string? name;
+    string name;
     int? age;
 |};
 
@@ -23,9 +23,9 @@ public function main() returns error? {
     // Creates a new client with the backend URL.
     graphql:Client graphqlClient = check new ("localhost:9090/graphql");
 
-    string document = "{ profile { name, age } }";
+    string document = "{ profile(id: 1) { name, age } }";
     ProfileResponse response = check graphqlClient->execute(document);
-    
+
     // Access the data from the response.
     io:println(response.data);
 
--- a/examples/graphql-service-error-handling/graphql_service_error_handling.bal
+++ b/examples/graphql-service-error-handling/graphql_service_error_handling.bal
@@ -29,7 +29,7 @@ service class Profile {
     // can be `null`. Hence, when an error returned from this method,
     resource function get age() returns int|error? {
         if self.id == 1 {
-            // return a mock error
+            // Returns a mock error.
             return error("Error occurred while retrieving age");
         }
         return self.age;
--- a/examples/graphql-service-error-handling/graphql_service_error_handling.bal
+++ b/examples/graphql-service-error-handling/graphql_service_error_handling.bal
@@ -1,0 +1,48 @@
+import ballerina/graphql;
+
+service class Profile {
+    private final int id;
+    private final string name;
+    private final int age;
+
+    function init(int id, string name, int age) {
+        self.id = id;
+        self.name = name;
+        self.age = age;
+    }
+
+    // This resource method does not have `nil` as a possible return type, which means the
+    // corresponding GraphQL field type is wrapped by GraphQL `NON_NULL` type. Therefore, if this
+    // method returns an `error`, the `name` value will become `null` and since it is a `NON_NULL`
+    // field, the value will be propagated to the upper level. This means the corresponding
+    // `profile` field value will become `null`.
+    resource function get name() returns string|error {
+        if self.id == 2 {
+            // return a mock error
+            return error("Error occurred while retrieving name");
+        }
+        return self.name;
+    }
+
+    // This resource method has `nil` as a possible return type, which means the corresponding
+    // GraphQL field type is nullable. Therefore, if this field returns an error, the field value
+    // can be `null`. Hence, when an error returned from this method,
+    resource function get age() returns int|error? {
+        if self.id == 1 {
+            // return a mock error
+            return error("Error occurred while retrieving age");
+        }
+        return self.age;
+    }
+}
+
+service /graphql on new graphql:Listener(9090) {
+
+    // This resource method returns only the `Profile` type, which means the field type is wrapped
+    // by GraphQL `NON_NULL` type. Therefore, if the `null` value is propagated to the `profile`
+    // field, (by returning an `error` from the `name` field in the `Profile` object), it will be
+    // propagated further, making the `data` field of the response `null`.
+    resource function get profile(int id) returns Profile {
+        return new (id, "Walter White", 50);
+    }
+}
--- a/examples/graphql-service-error-handling/graphql_service_error_handling.bal
+++ b/examples/graphql-service-error-handling/graphql_service_error_handling.bal
@@ -12,13 +12,13 @@ service class Profile {
     }
 
     // This resource method does not have `nil` as a possible return type, which means the
-    // corresponding GraphQL field type is wrapped by GraphQL `NON_NULL` type. Therefore, if this
+    // corresponding GraphQL field type is wrapped by the GraphQL `NON_NULL` type. Therefore, if this
     // method returns an `error`, the `name` value will become `null` and since it is a `NON_NULL`
     // field, the value will be propagated to the upper level. This means the corresponding
     // `profile` field value will become `null`.
     resource function get name() returns string|error {
         if self.id == 2 {
-            // return a mock error
+            // Returns a mock error.
             return error("Error occurred while retrieving name");
         }
         return self.name;
@@ -40,8 +40,8 @@ service /graphql on new graphql:Listener(9090) {
 
     // This resource method returns only the `Profile` type, which means the field type is wrapped
     // by GraphQL `NON_NULL` type. Therefore, if the `null` value is propagated to the `profile`
-    // field, (by returning an `error` from the `name` field in the `Profile` object), it will be
-    // propagated further, making the `data` field of the response `null`.
+    // field (by returning an `error` from the `name` field in the `Profile` object), it will be
+    // propagated further making the `data` field of the response `null`.
     resource function get profile(int id) returns Profile {
         return new (id, "Walter White", 50);
     }
--- a/examples/graphql-service-error-handling/graphql_service_error_handling.bal
+++ b/examples/graphql-service-error-handling/graphql_service_error_handling.bal
@@ -26,7 +26,7 @@ service class Profile {
 
     // This resource method has `nil` as a possible return type, which means the corresponding
     // GraphQL field type is nullable. Therefore, if this field returns an error, the field value
-    // can be `null`. Hence, when an error returned from this method,
+    // can be `null`.
     resource function get age() returns int|error? {
         if self.id == 1 {
             // Returns a mock error.
--- a/examples/graphql-service-error-handling/graphql_service_error_handling.md
+++ b/examples/graphql-service-error-handling/graphql_service_error_handling.md
@@ -26,7 +26,7 @@ To send the document, execute the following cURL command in a separate terminal.
 
 ::: out graphql_service_error_handling.2.client.out :::
 
-Check the response to see the `data` field is set to null due to propagating the `null` value because the `name` field and the `profile` field are of `NON_NULL` type.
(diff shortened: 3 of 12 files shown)


examples/graphql-client-handle-partial-response/graphql_client_handle_partial_response.bal
```diff
@@ -13,19 +13,19 @@ type ProfileResponse record {|
 |};
 
 // The following record type defines the shape of the response from a GraphQL service which allows
```
praneesha, line 15:
```suggestion
// The following record type defines the shape of the response from a GraphQL service, which allows
```

examples/graphql-service-error-handling/graphql_service_error_handling.bal
```diff
@@ -0,0 +1,48 @@
...
+    private final int age;
+
+    function init(int id, string name, int age) {
+        self.id = id;
+        self.name = name;
+        self.age = age;
+    }
+
+    // This resource method does not have `nil` as a possible return type, which means the
+    // corresponding GraphQL field type is wrapped by GraphQL `NON_NULL` type. Therefore, if this
```
praneesha, line 15:
```suggestion
    // corresponding GraphQL field type is wrapped by the GraphQL `NON_NULL` type. Therefore, if this
```

examples/graphql-service-error-handling/graphql_service_error_handling.bal
```diff
@@ -0,0 +1,48 @@
...
+    // method returns an `error`, the `name` value will become `null` and since it is a `NON_NULL`
+    // field, the value will be propagated to the upper level. This means the corresponding
+    // `profile` field value will become `null`.
+    resource function get name() returns string|error {
+        if self.id == 2 {
+            // return a mock error
```
praneesha, line 21:
```suggestion
            // Returns a mock error.
```

examples/graphql-service-error-handling/graphql_service_error_handling.bal
```diff
@@ -0,0 +1,48 @@
...
+        }
+        return self.name;
+    }
+
+    // This resource method has `nil` as a possible return type, which means the corresponding
+    // GraphQL field type is nullable. Therefore, if this field returns an error, the field value
+    // can be `null`. Hence, when an error returned from this method,
+    resource function get age() returns int|error? {
+        if self.id == 1 {
+            // return a mock error
```
praneesha, line 32:
```suggestion
            // it returns a mock error.
```

examples/graphql-service-error-handling/graphql_service_error_handling.bal
```diff
@@ -0,0 +1,48 @@
...
+        }
+        return self.age;
+    }
+}
+
+service /graphql on new graphql:Listener(9090) {
+
+    // This resource method returns only the `Profile` type, which means the field type is wrapped
+    // by GraphQL `NON_NULL` type. Therefore, if the `null` value is propagated to the `profile`
+    // field, (by returning an `error` from the `name` field in the `Profile` object), it will be
```
praneesha, line 43:
```suggestion
    // field (by returning an `error` from the `name` field in the `Profile` object), it will be
```

examples/graphql-service-error-handling/graphql_service_error_handling.bal
```diff
@@ -0,0 +1,48 @@
...
+    // propagated further, making the `data` field of the response `null`.
```
praneesha, line 44:
```suggestion
    // propagated further making the `data` field of the response `null`.
```

examples/graphql-service-error-handling/graphql_service_error_handling.md
```diff
@@ -0,0 +1,35 @@
+# GraphQL service - Error handling
+
+The Ballerina `graphql` module allows returning `error`s from the `resource` and `remote` methods used as GraphQL object fields. If a field returns an error, it will be added in the GraphQL response under the `errors` field, while the field value is set to `null`. In this case, if the GraphQL field type is `NON_NULL`, the `null` value is propagated to the upper levels, until a `null` value is allowed as the field value. This might cause the whole `data` field to be `null` in some cases. Alternatively, the field type can include nil, so the field value can be `null`. In that case, the `null` value is not propagated to the upper levels in the response.
```
praneesha, line 3:
```suggestion
The Ballerina `graphql` module allows returning `error`s from the `resource` and `remote` methods used as GraphQL object fields. If a field returns an error, it will be added in the GraphQL response under the `errors` field while the field value is set to `null`. In this case, if the GraphQL field type is `NON_NULL`, the `null` value is propagated to the upper levels until a `null` value is allowed as the field value. This might cause the whole `data` field to be `null` in some cases. Alternatively, the field type can include nil, so the field value can be `null`. In that case, the `null` value is not propagated to the upper levels in the response.
```

examples/graphql-service-error-handling/graphql_service_error_handling.md
```diff
@@ -0,0 +1,35 @@
...
+
+::: code graphql_service_error_handling.bal :::
+
+Run the service by executing the following command.
```
praneesha, line 7:
```suggestion
Run the service by executing the following command.
```
Shall we add these as numbered steps?

examples/graphql-service-error-handling/graphql_service_error_handling.md
```diff
@@ -0,0 +1,35 @@
...
+
+::: out graphql_service_error_handling.server.out :::
+
+Send the following document to the GraphQL endpoint to test the service.
+
+::: code graphql_service_error_handling.1.graphql :::
+
+To send the document, use the following cURL command in a separate terminal.
```
praneesha, line 15:
```suggestion
To send the document, execute the following cURL command in a separate terminal.
```

examples/graphql-service-error-handling/graphql_service_error_handling.md
```diff
@@ -0,0 +1,35 @@
...
+
+::: out graphql_service_error_handling.1.client.out :::
+
+Check the response to see how both the `errors` and the `data` fields are present in the response.
+
+Now, send the following document to the GraphQL endpoint.
+
+::: code graphql_service_error_handling.2.graphql :::
+
+To send the document, use the following cURL command in a separate terminal.
```
praneesha, line 25:
```suggestion
To send the document, execute the following cURL command in a separate terminal.
```

examples/graphql-service-error-handling/graphql_service_error_handling.md
```diff
@@ -0,0 +1,35 @@
...
+
+::: out graphql_service_error_handling.2.client.out :::
+
+Check the response to see the `data` field is set to null due to propagating the `null` value because the `name` field and the `profile` field are of `NON_NULL` type.
```
praneesha, line 29:
```suggestion
Check the response to see if the `data` field is set to null due to propagating the `null` value because the `name` field and the `profile` field are of the `NON_NULL` type.
```

examples/graphql-service-error-handling/graphql_service_error_handling.bal
ThisaruGuruge: Shouldn't this be 
```suggestion
            // Returns a mock error
```

to make it similar to the comment above?

examples/graphql-service-error-handling/graphql_service_error_handling.md
ThisaruGuruge: We did not add numbers to the BBEs similar to this. Wouldn't adding numbers make this out-of-convention?

examples/graphql-service-error-handling/graphql_service_error_handling.md
ThisaruGuruge: This was suggested by Grammarly, but I specifically ignored it because here we want the user to check the response, in which the `data` field should be `null`. I'm not sure if the `see if` conveys the message correctly.

examples/graphql-service-error-handling/graphql_service_error_handling.bal
praneesha: This seems to be a continuation of the previous comment.

examples/graphql-service-error-handling/graphql_service_error_handling.md
praneesha: The best practice in technical documentation is to add sequential instructions as numbered bullet points. We can adhere to this as now we have the MD format.

examples/graphql-service-error-handling/graphql_service_error_handling.md
praneesha: The Grammarly suggestion is correct.

examples/graphql-service-error-handling/graphql_service_error_handling.bal
ThisaruGuruge: Updated.

examples/graphql-service-error-handling/graphql_service_error_handling.md
ThisaruGuruge: I think the better wording should be `see how`, instead of `see if`. Would that be okay?

examples/graphql-service-error-handling/graphql_service_error_handling.md
ThisaruGuruge: In that case, we might have to do this in every BBE. @shafreenAnfar WDYT?

examples/graphql-service-error-handling/graphql_service_error_handling.md
praneesha: +1 to that.

examples/graphql-service-error-handling/graphql_service_error_handling.metatags
```diff
@@ -0,0 +1,2 @@
+description: A GraphQL service endpoint written in Ballerina
+keywords: ballerina, ballerina by example, bbe, graphql, graphql union type
```
MohamedSabthar, line 2:
```suggestion
keywords: ballerina, ballerina by example, bbe, graphql, error handling
```

examples/graphql-service-error-handling/graphql_service_error_handling.md
```diff
@@ -0,0 +1,35 @@
+# GraphQL service - Error handling
+
+The Ballerina `graphql` module allows returning `error`s from the `resource` and `remote` methods used as GraphQL object fields. If a field returns an error, it will be added in the GraphQL response under the `errors` field while the field value is set to `null`. In this case, if the GraphQL field type is `NON_NULL`, the `null` value is propagated to the upper levels until a `null` value is allowed as the field value. This might cause the whole `data` field to be `null` in some cases. Alternatively, the field type can include nil, so the field value can be `null`. In that case, the `null` value is not propagated to the upper levels in the response.
```
MohamedSabthar, line 3:
```suggestion
The Ballerina `graphql` module allows returning `error`s from the `resource` and `remote` methods used as GraphQL object fields. If a field returns an error, it will be added in the GraphQL response under the `errors` field while the field value is set to `null`. In this case, if the GraphQL field type is `NON_NULL`, the `null` value is propagated to the upper levels until a `null` value is allowed as the field value. This might cause the whole `data` field to be `null` in some cases. Alternatively, the field type can include `nil`, so the field value can be `null`. In that case, the `null` value is not propagated to the upper levels in the response.
```

MohamedSabthar (approved):

examples/graphql-service-error-handling/graphql_service_error_handling.md
ThisaruGuruge: Updated.

examples/graphql-service-error-handling/graphql_service_error_handling.md
shafreenAnfar: Lets not change the convention now. Can consider it for a later revamp.

examples/graphql-service-error-handling/graphql_service_error_handling.md
```diff
@@ -0,0 +1,35 @@
+# GraphQL service - Error handling
+
+The Ballerina `graphql` module allows returning `error`s from the `resource` and `remote` methods used as GraphQL object fields. If a field returns an error, it will be added in the GraphQL response under the `errors` field while the field value is set to `null`. In this case, if the GraphQL field type is `NON_NULL`, the `null` value is propagated to the upper levels until a `null` value is allowed as the field value. This might cause the whole `data` field to be `null` in some cases. Alternatively, the field type can include `nil`, so the field value can be `null`. In that case, the `null` value is not propagated to the upper levels in the response.
```
shafreenAnfar, line 3: How about something like the below with minor alterations

The Ballerina `graphql` module allows returning `error`s from the `resource` and `remote` methods. When the method returns an error, it is added in the GraphQL response under the `errors` field whereas the field value is set to `null` under the `data` field. However, if the remote method’s return type does not also include `nil`, the GraphQL field type becomes  `NON_NULL`. Therefore,  the `null` value is propagated to the upper levels until a `null` value is allowed as the field value. This might cause the whole `data`  field in the GraphQL response to be `null` in some cases. To avoid this cascading behavior, the return type must include `nil`.

ThisaruGuruge: I did not add the `must return nil` part because that might send the wrong idea that `when you return an error, you must include nil`, which is not true. Do we need that sentence?

ThisaruGuruge: Maybe `if this cascading behavior needs to be avoided, the return type should include nil`?

shafreenAnfar: okay, How about

To avoid this cascading behavior, the return type should include nil as appropriate.

ThisaruGuruge: +1

ThisaruGuruge: Updated.

shafreenAnfar (approved):

PR #3937, opened 2023-01-05
Fix issues related to list `BBE`s

rdulmina:
1. Update incorrect `.bal` and `.out` references in the list related BBE `.md` files
2. Remove incorrect line breaks in metatags

### DIFF
--- a/examples/list-equality/list_equality.md
+++ b/examples/list-equality/list_equality.md
@@ -1,6 +1,6 @@
 # Filler values of a list
 
-We can use `==` and `!=` on lists to check deep equality of two lists: two lists are deep equal if they have the same members in the same order. Deep equality only works for `anydata` lists. `===` and `!==` check for the exact equality which matches the references of the lists.
+You can use `==` and `!=` on lists to check the deep equality of two lists: two lists are deep equal if they have the same members in the same order. Deep equality only works for `anydata` lists. `===` and `!==` check for the exact equality, which matches the references of the lists.
 
 ::: code list_equality.bal :::
 
--- a/examples/list-equality/list_equality.md
+++ b/examples/list-equality/list_equality.md
@@ -2,9 +2,9 @@
 
 We can use `==` and `!=` on lists to check deep equality of two lists: two lists are deep equal if they have the same members in the same order. Deep equality only works for `anydata` lists. `===` and `!==` check for the exact equality which matches the references of the lists.
 
-::: code updating_a_list.bal :::
+::: code list_equality.bal :::
 
-::: out updating_a_list.out :::
+::: out list_equality.out :::
 
 ## Related links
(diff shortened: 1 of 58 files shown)


examples/list-equality/list_equality.md
```diff
@@ -2,9 +2,9 @@
 
 We can use `==` and `!=` on lists to check deep equality of two lists: two lists are deep equal if they have the same members in the same order. Deep equality only works for `anydata` lists. `===` and `!==` check for the exact equality which matches the references of the lists.
```
praneesha, line 3:
```suggestion
You can use `==` and `!=` on lists to check the deep equality of two lists: two lists are deep equal if they have the same members in the same order. Deep equality only works for `anydata` lists. `===` and `!==` check for the exact equality, which matches the references of the lists.
```

examples/list-equality/list_equality.md
```diff
@@ -2,9 +2,9 @@
...
 
-::: code updating_a_list.bal :::
+::: code list_equality.bal :::
 
-::: out updating_a_list.out :::
+::: out list_equality.out :::
 
 ## Related links
 - [Tuples - Ballerina by example](https://ballerina.io/learn/by-example/tuples)
```
praneesha, line 10:
```suggestion
- [Tuples](https://ballerina.io/learn/by-example/tuples)
```

examples/list-subtyping/list_subtyping.md
```diff
@@ -2,9 +2,9 @@
 
 An array with a length is a subtype of an array of same member type descriptor without a length. Two arrays with the same length or  without a length is a subtype of the other if the member type descriptor is a subtype of the other. Tuple type `T` is a subtype of  type `T1` if the set of values allowed in `T` is a sub set of values allowed in `T1`
```
praneesha, line 3:
```suggestion
An array with a length is a subtype of an array of the same member type descriptor without a length. Two arrays with the same length or without a length are a subtype of the other if the member type descriptor is a subtype of the other. Tuple type `T` is a subtype of  type `T1` if the set of values allowed in `T` is a subset of the values allowed in `T1`
```

examples/list-equality/list_equality.md
rdulmina: IMO its better to keep the existing one to align with the `- Specification` and `- API documentation` suffixes in the existing BBEs

gimantha: @praneesha @hasithaa shouldnt we have any indication of what type of related link this is? for example, BBE, Spec orGuide ?

rdulmina: Please refer https://ballerina.io/learn/by-example/websocket-service-payload-constraint-validation/

hasithaa (approved):

PR #3985, opened 2023-01-17
Sync 2201.3.x branch with changes in bbe-iteration3 branch

chamil321:


examples/http-caller/http_caller.md
```diff
@@ -1,4 +1,4 @@
-# HTTP caller
+# HTTP caller object
```
shafreenAnfar, line 1: HTTP service - Caller object

PR #3995, opened 2023-01-18
Fix JSON BBE

ushirask:


examples/access-optional-json-elements/access_optional_json_elements.out
```diff
@@ -1,2 +1,2 @@
 $ bal run working_directly_with_json.bal
```
gimantha, line 1: the file name should be changed to `access_optional_json_elements.bal` right

ushirask: Fixed.

gimantha (approved):

PR #4122, opened 2023-02-23
Add extra installer for MACOS-ARM processors

### LINKED ISSUE #3997: Provide a mac installer with JRE supports M1
**Description:**
Currently Ballerina has installer with x64 JRE inbuilt. This is causing issues in GraalVM  (https://github.com/ballerina-platform/ballerina-lang/issues/39003). In that case we need to provide a mac installer with JRE supports M1

- [ ] Archive an installer for macos ARM processors
- [ ] Update tool support for new macos platform
- [ ] Redesign website download page to include new installer

udda1996:
## Purpose
partially fixes #3997


installers/mac-arm/build-ballerina-macos-arm-x64.sh
```diff
@@ -0,0 +1,160 @@
+#!/bin/bash
+
+function printUsage() {
```
keizer619, line 3: Can't we use the same script and generate 2 installers by switching zip source files? 
We may have to parameterize few things to support both

sm1990 (approved):

PR #4434, opened 2023-05-17
Add BBE for group by and collect

KavinduZoysa:
## Purpose
$subject


KavinduZoysa:
This is covered via https://github.com/ballerina-platform/ballerina-distribution/pull/4686

PR #4507, opened 2023-05-31
update nexus for lang,azure,docker,kubenetes and aws

SasinduDilshara:
update nexus for lang and azure

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -117,7 +117,7 @@
         <repository>
             <id>wso2.docker.stage</id>
             <name>WSO2 Stage Repository</name>
-            <url>https://maven.wso2.org/nexus/content/repositories/orgballerinaxdocker-1368/</url>
+            <url>https://maven.wso2.org/nexus/content/repositories/orgballerinaxdocker-1369/</url>
             <releases>
                 <enabled>true</enabled>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -304,9 +304,9 @@
         <ballerina.version>1.2.39</ballerina.version>
         <ballerina.jre.artifacts.version>2.0.1</ballerina.jre.artifacts.version>
         <project.scm.id>ballerina-scm-server</project.scm.id>
-        <docker.version>1.2.39</docker.version>
-        <kubernetes.version>1.2.39</kubernetes.version>
-        <awslambda.version>1.2.39</awslambda.version>
+        <docker.version>1.2.38</docker.version>
```
malinthar, line 307: Why are we downgrading these? Didn't we release them

SasinduDilshara: it is just for this PR will update when the ara available on nexus

malinthar: We have to release them first and then update the distribution. That way it would be easier right? No need to create 2 PRs

SasinduDilshara: Done

chiranSachintha (approved):

malinthar (approved):

PR #4514, opened 2023-06-01
Add EDI BBEs for 2201.6.x

chathurace:


examples/edi-to-record/edi_to_record.bal
```diff
@@ -0,0 +1,12 @@
+import ballerina/io;
+import edi_to_record.sorder;
```
anupama-pathirage, line 2: Practice is to import the modules from the same package first.

anupama-pathirage (approved):

PR #4539, opened 2023-06-05
Sync master with 2201.6.x

SasinduDilshara:
Sync master with 2201.6.x


gradle.properties
```diff
@@ -8,85 +8,86 @@ ballerinaJreVersion=1.1.0
 dependencyJREVersion=jdk-11.0.18+10-jre
 specVersion=2022R4
 slf4jVersion=1.7.30
-balstdlibBranch=master
+balstdlibBranch=2201.6.x
```
NipunaMadhushan, line 11: this should be left as `master`

SasinduDilshara: will update

PR #4648, opened 2023-07-01
Update publish release and sign installers yaml

Miranlfk:
## Purpose
$title

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -73,7 +73,7 @@ jobs:
           devCentralToken: ${{ secrets.BALLERINA_DEV_CENTRAL_ACCESS_TOKEN }}
           ballerinaBotWorkflow: $ {{ secrets.BALLERINA_BOT_WORKFLOW }}
         run: |
-          ./gradlew build -Pversion=${VERSION}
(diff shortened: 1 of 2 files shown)


Miranlfk:
Added, a job to retrieve and pass `branch_name` which will be utilized in `cosign verify-blob` command text. Used `sed` to remove first 10 lines used for optimization in md file. 

**Note:** 
The distribution repo retrieves the `verify-ballerina-artifacts.md` from `dev-website` repo which is used on the release page in GitHub. Only the saved version of this md file during the workflow run is edited with changes. the original `verify-ballerina-artifacts.md` file is not affected.

.github/workflows/publish-release.yml
```diff
@@ -73,7 +73,7 @@ jobs:
           devCentralToken: ${{ secrets.BALLERINA_DEV_CENTRAL_ACCESS_TOKEN }}
           ballerinaBotWorkflow: $ {{ secrets.BALLERINA_BOT_WORKFLOW }}
         run: |
-          ./gradlew build -Pversion=${VERSION}
+          ./gradlew build -Pversion=${VERSION} --exclude-task project-api-tests:test
```
udda1996, line 76:
```suggestion
          ./gradlew build -Pversion=${VERSION}
```

udda1996 (approved):

PR #4713, opened 2023-07-14
[2201.8.x] Update const BBE

prakanth97:
## Purpose
Add list constant and map constant examples

### DIFF
--- a/examples/const-and-final/const_and_final.bal
+++ b/examples/const-and-final/const_and_final.bal
@@ -1,10 +1,11 @@
 import ballerina/io;
 
+const int MAX_VALUE = 1000;
+
 // Constants can be defined without the type. Then, the type is inferred from the right-hand side.
-const MAX_VALUE = 1000;
 const URL = "https://ballerina.io";
 
-// Map and list can be declared as constant.
+// Mapping and list constants can also be defined.
 const HTTP_OK = {httpCode: 200, message: "OK"};
 const ERROR_CODES = [200, 202, 400];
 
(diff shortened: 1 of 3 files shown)


examples/const-and-final/const_and_final.bal
```diff
@@ -4,12 +4,18 @@ import ballerina/io;
 const MAX_VALUE = 1000;
```
MaryamZi, line 4: Not introduced by this PR, but do we know why both don't have the type here? I think one can have the type.

examples/const-and-final/const_and_final.bal
```diff
@@ -4,12 +4,18 @@ import ballerina/io;
...
 const URL = "https://ballerina.io";
 
+// Map and list can be declared as constant.
```
MaryamZi, line 7:
```suggestion
// Mapping and list constants can also be defined.
```

examples/const-and-final/const_and_final.bal
prakanth97: +1

prakanth97: Updated [d250c9f](https://github.com/ballerina-platform/ballerina-distribution/pull/4713/commits/d250c9f661229b921af0f6dc9bea2f0b491595fb)

hasithaa (approved):

rdulmina:
Closing this PR since these changes are sent to the master branch and we are recreating `2201.8.x` from that master later.

PR #5191, opened 2024-02-12
Fix list subtyping BBE output

lochana-chathura:

### DIFF
--- a/examples/list-subtyping/list_subtyping.bal
+++ b/examples/list-subtyping/list_subtyping.bal
@@ -2,6 +2,8 @@ import ballerina/io;
 
 public function main() {
     int[3] numbers = [144, 232, 322];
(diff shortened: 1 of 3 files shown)


examples/list-subtyping/list_subtyping.bal
```diff
@@ -2,6 +2,8 @@ import ballerina/io;
 
 public function main() {
     int[3] numbers = [144, 232, 322];
+    // `numbers` will be a subtype of `int[]`
```
MaryamZi, line 5: Not specific to your change, but would it make more sense to say
```suggestion
    // `numbers`, of type `int[3]`, will be a subtype of `int[]`
```

MaryamZi (dismissed):

MaryamZi (approved):

PR #5195, opened 2024-02-13
Update GraphQL and Dev Tools Versions

DimuthuMadushan:


ThisaruGuruge:
Is this synced correctly? We already merged the BBE PR with the GraphQL module version bump, right?

PR #5295, opened 2024-04-08
Add BBEs for Xmldata Module

prakanth97:
## Purpose

- Added BBE for convert XML to record
- Added BBE for convert XML to record with projection

### DIFF
--- a/examples/xml-to-record-with-projection/xml_to_record_with_projection.bal
+++ b/examples/xml-to-record-with-projection/xml_to_record_with_projection.bal
@@ -7,26 +7,26 @@ type Book record {|
     string author;
 |};
 
-public function main() returns error? {
-    xml xmlData = xml `
-        <book>
-            <name>Clean Code</name>
-            <author>Robert C. Martin</author>
-            <year>2008</year>
-            <publisher>Prentice Hall</publisher>
-        </book>`;
+xml xmlData = xml `
+<book>
+    <name>Clean Code</name>
+    <author>Robert C. Martin</author>
+    <year>2008</year>
+    <publisher>Prentice Hall</publisher>
+</book>`;
 
+string xmlStr = string `
+<book>
+    <name>Clean Code</name>
+    <author>Robert C. Martin</author>
+    <year>2008</year>
+    <publisher>Prentice Hall</publisher>
+</book>`;
+
+public function main() returns error? {
     // Based on the expected type, it selectively converts the XML data to the record type.
     Book book = check xmldata:parseAsType(xmlData);
     io:println(book);
-
-    string xmlStr = string `
-        <book>
-            <name>Clean Code</name>
-            <author>Robert C. Martin</author>
-            <year>2008</year>
-            <publisher>Prentice Hall</publisher>
-        </book>`;
 
     // Based on the expected type, it selectively converts the XML string to the record type.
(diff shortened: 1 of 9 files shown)


examples/xml-to-record/xml_to_record.bal
```diff
@@ -0,0 +1,65 @@
...
+    // Construct stream of byte blocks.
+    stream<byte[], error?> byteBlockStream = new(new XmlByteBlock());
+    // Convert the XML byte block stream to a record type.
+    Book book4 = check xmldata:parseStream(byteBlockStream);
+    io:println(book4);
+}
+
+// Defines a class called `XmlByteBlock`, which implements the `next()` method.
+// This will be invoked when the `next()` method of the stream gets invoked.
+class XmlByteBlock {
```
LakshanWeerasinghe, line 45: I think it would be better if we can give a more generic byte block stream implementation.
```ballerina
class ByteBlockGenerator {
    private int index = 0;
    private final byte[] byteArr;
    private final int arraySize;

    public function init(string data) {
        self.byteArr = data.toBytes();
        self.arraySize = self.byteArr.length();
    }

    public isolated function next() returns record {|byte[] value;|}|error? {
        if self.index >= self.arraySize {
            return;
        }
        int startIndex = self.index;
        self.index = startIndex + 3 >= self.arraySize ? self.arraySize - 1 : startIndex + 3;
        return {value: self.byteArr.slice(startIndex, self.index)};
    }
}

```

examples/xml-to-record/xml_to_record.bal
```diff
@@ -0,0 +1,63 @@
...
+    // Convert the XML string to a record type.
+    Book book2 = check xmldata:parseString(xmlStr);
+    io:println(book2);
+
+    byte[] xmlByteArr = xmlStr.toBytes();
+    // Convert the XML byte array to a record type.
+    Book book3 = check xmldata:parseBytes(xmlByteArr);
+    io:println(book3);
+
+    // Construct stream of byte blocks.
```
LakshanWeerasinghe, line 36:
```suggestion
```

prakanth97: Updated

LakshanWeerasinghe (dismissed):

examples/xml-to-record-with-projection/xml_to_record_with_projection.bal
```diff
@@ -0,0 +1,34 @@
...
+            <author>Robert C. Martin</author>
+            <year>2008</year>
+            <publisher>Prentice Hall</publisher>
+        </book>`;
+
+    // Based on the expected type, it selectively converts the XML data to the record type.
+    Book book = check xmldata:parseAsType(xmlData);
+    io:println(book);
+
+    string xmlStr = string `
```
SasinduDilshara, line 23: imo, define this in module level is more clear. wdyt?

prakanth97: Updated https://github.com/ballerina-platform/ballerina-distribution/pull/5295/commits/07acb43a5a2aa3ce5a5e0c72d5efe81e0ff0cd96

SasinduDilshara (approved):

hasithaa (approved):

PR #5553, opened 2024-08-08
Add LDAP module into the Ballerina distribution

ayeshLK:


ThisaruGuruge:
Shall we send this change to the master branch as well?

ayeshLK:
> Shall we send this change to the master branch as well?

will do that.

PR #5594, opened 2024-08-21
[2201.10.0-stage] Add client invocation tip for SSE server

TharmiganK:

### DIFF
--- a/examples/http-sse-service/http_sse_service.md
+++ b/examples/http-sse-service/http_sse_service.md
@@ -12,7 +12,7 @@ Invoke the service by executing the following cURL command in a new terminal.
 
 ::: out http_sse_service.client.out :::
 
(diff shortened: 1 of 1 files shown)


examples/http-sse-service/http_sse_service.md
```diff
@@ -12,6 +12,8 @@ Invoke the service by executing the following cURL command in a new terminal.
 
 ::: out http_sse_service.client.out :::
 
+>**Tip:** You can invoke the above service via the [Client server-sent events](learn/by-example/http-sse-client/) example.
```
TharmiganK, line 15:
```suggestion
>**Tip:** You can invoke the above service via the [Client server-sent events](/learn/by-example/http-sse-client/) example.
```

PR #5632, opened 2024-09-10
Update Ballerina By Examples to include references to the `binary data` BBE links

SasinduDilshara:
Update Ballerina By Examples to include references to the binary data and list equality BBE links

Fixes https://github.com/ballerina-platform/ballerina-dev-website/issues/9385

PR #5656, opened 2024-09-18
Update Ballerina By Examples to include references to the binary data BBE links

SasinduDilshara:
related to https://github.com/ballerina-platform/ballerina-distribution/pull/5632


NipunaRanasinghe (commented):
@SasinduDilshara IMO we can update the `security_crypto` BBE too with the same reference, as its also having quite a few binary data usages. wdyt?

NipunaRanasinghe (approved):

PR #5657, opened 2024-09-18
Update binding patterns BBEs

poorna2152:
## Purpose
Closes https://github.com/ballerina-platform/ballerina-dev-website/issues/9305

### DIFF
--- a/examples/list-binding-pattern-in-match-statement/list_binding_pattern_in_match_statement.bal
+++ b/examples/list-binding-pattern-in-match-statement/list_binding_pattern_in_match_statement.bal
@@ -20,9 +20,32 @@ function matchCommand(any commands) {
     }
 }
 
+function matchWithMatchGuard(any lst) {
+    match lst {
+        // The list binding pattern below binds lists that contain two members
+        // of the types `string` and `int` respectively.
+        var [s, i] if s is string && i is int => {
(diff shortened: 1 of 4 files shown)


examples/list-binding-pattern-in-match-statement/list_binding_pattern_in_match_statement.bal
```diff
@@ -20,9 +20,32 @@ function matchCommand(any commands) {
     }
 }
 
+function matchWithMatchGuard(any lst) {
+    match lst {
+        // The list binding pattern below binds lists that contain two members
+        // of the types `string` and `int` respectively.
+        var [s, i] if s is string && i is int => {
```
gimantha, line 27: Wouldnt it be better to have the type guard as a separate BBE? @MaryamZi WDYT?

MaryamZi: Yeah, if we are adding this many. If not, we can check if we can add just one clause with a guard to matchCommand.

MaryamZi: I prefer the latter btw. We don't need a lot of examples IMO.

PR #5664, opened 2024-09-20
Update on conflict clause `BBE` with more real-world sample

rdulmina:
## Purpose
Fix https://github.com/ballerina-platform/ballerina-dev-website/issues/9288


examples/on-conflict-clause/on_conflict_clause.bal
```diff
@@ -7,32 +7,53 @@ type Student record {|
...
-    // The value `100` of the key `John` will be replaced by the second occurrence value `200`.
-    map<int> lastRoundScore = map from var student in students
-                              select [student.name, student.score]
-                              on conflict ();
-
-    io:println(lastRoundScore);
-
-    // The result of the following will be an error since the key `1` is duplicated.
+    // The result of the following will be an error since the key `1` is duplicated
+    // and `isSafeReplace()` function returns an error for the name `John`
```
SasinduDilshara, line 25:
```suggestion
    // The result of the following will be an error since the key `1` is duplicated
    // and `isSafeReplace()` function returns an error for the name `John`.
```

examples/on-conflict-clause/on_conflict_clause.bal
```diff
@@ -7,32 +7,53 @@ type Student record {|
...
-
     table<Student> students = table [
             {id: 1, name: "John", score: 100},
             {id: 2, name: "Jane", score: 150},
             {id: 1, name: "John", score: 200}
         ];
 
-    // The result of the following will be an error since the key `John` is duplicated.
+    // The result of the following will be an error since the key `John` is duplicated
+    // and `isSafeReplace()` function returns an error for the name `John`
```
SasinduDilshara, line 17:
```suggestion
    // The result of the following will be an error since the key `John` is duplicated
    // and `isSafeReplace()` function returns an error for the name `John`.
```

examples/on-conflict-clause/on_conflict_clause.bal
```diff
@@ -7,32 +7,53 @@ type Student record {|
...
+            {id: 1, name: "John", score: 100},
+            {id: 2, name: "Mike", score: 150},
+            {id: 2, name: "Mike", score: 200}
+        ];
+
+    // The value `100` of the key `Mike` will be replaced by the second occurrence value `200`
+    // since `isSafeReplace()` function returns `null` for the name 'Mike`
+    map<int>|error lastRoundScore = map from var student in students2
+                                    select [student.name, student.score]
+                                    on conflict isSafeReplace(student.name);
```
gimantha, line 42: why do we need a function here? shall we simply return an error for simplicity?

rdulmina: The intention is to demonstrate conflict handling given a frame variable which is a more realistic use case rather than just returning the error or null

examples/on-conflict-clause/on_conflict_clause.bal
```diff
@@ -7,32 +7,53 @@ type Student record {|
...
-    error onConflictError = error("Key Conflict", message = "record with same key exists.");
-
     table<Student> students = table [
             {id: 1, name: "John", score: 100},
             {id: 2, name: "Jane", score: 150},
             {id: 1, name: "John", score: 200}
         ];
 
-    // The result of the following will be an error since the key `John` is duplicated.
+    // The result of the following will be an error since the key `John` is duplicated
```
gimantha, line 16: We need to say why the result is error. It is because we have specifically used `on conflict` clause

rdulmina: Updated the description to explain that since it is the common behavior

examples/on-conflict-clause/on_conflict_clause.bal
```diff
@@ -7,32 +7,53 @@ type Student record {|
...
+    // always returns `()`.
+    map<int> lastRoundScore2 = map from var student in students2
+                               select [student.name, student.score]
+                               on conflict ();
+
+    io:println(lastRoundScore2);
+}
+
+function isSafeReplace(string name) returns error? {
+    if name == "John" {
```
MaryamZi, line 56: IMO, a check like this still doesn't make it like a real world example? What if we just have two examples, one where we just use an error constructor in the on conflict clause to return an error (or call a function that logs at error level and returns an error) and the other call a function that just logs a warning saying the existing value is replaced on conflict and returns nil?

examples/on-conflict-clause/on_conflict_clause.md
```diff
@@ -2,7 +2,7 @@
 
 When constructing a map or a table with a key sequence using a query expression, there can be conflicting keys. The `on conflict` clause can be specified after the `select` clause to handle these cases.
 
-The syntax to write an `on conflict` clause is `on conflict expression`. The type of the expression should be `error?`. If the result of evaluating the expression is an error, then, the error will be the result. Otherwise, the old value is replaced by the new value.
+The `on conflict` clause uses the syntax `on conflict <expression>`, where the expression must be of type `error?`. If a conflicting key is found, and the expression evaluates to an error, that error becomes the result of the query expression. If it evaluates to `null` or the `on conflict` clause is absent, the old value is replaced by the new value.
```
MaryamZi, line 5:
```suggestion
The `on conflict` clause uses the syntax `on conflict <expression>`, where the expression must be of type `error?`. If a conflicting key is found, and the expression evaluates to an error, that error becomes the result of the query expression. If it evaluates to nil or the `on conflict` clause is absent, the old value is replaced by the new value.
```

examples/on-conflict-clause/on_conflict_clause.bal
rdulmina: IMO, we should demonstrate error handling given a frame variable. Param `name` in this case. Static error returning and logging are not ideal. Using a frame variable in error handling is the purpose of creating this issue earlier. Hence, I think the existing function is simpler and demonstrates the purpose

MaryamZi: We can just include `name` in the error/warning log or the error value? 

BBEs shouldn't be too complicated. We just need to make sure the idea is conveyed in a meaningful example.

rdulmina: Including the `name` in the log and always returning an error is not the ideal sample right? We must convey to the user that he can do some logic with the frame variables and decide whether to replace the value or give an error. Also, I believe the if condition complicates the BBE since it is very straightforward.

MaryamZi: > We must convey to the user that he can do some logic with the frame variables and decide whether to replace the value or give an error.

Isn't that conveyed via having a function being called with the variable as an argument? I think it's self-explanatory. 

> Including the `name` in the log and always returning an error is not the ideal sample right?

What we want to do IMO is explain that you can return error or nil and how the behaviour changes. 

I think the common case is that for all on conflict scenarios you either return error or you return nil. Rather than returning error in some cases and nil in others. 

If we really need an error? example, still checking for an arbitrary name like we've done here isn't real-world-like IMO. A better check would be to check for restricted names like `admin`, which still isn't ideal. Can we rethink the example?

rdulmina: +1 to change the name to `admin`. Why would you think it still isn't ideal? It conveys to the user that the function returns errors for some special cases depending on the frame variable.

MaryamZi: Because it's not something you'd even expect being in the collection to start with (i.e., not accessible to the end user).

rdulmina: The idea that we want to demonstrate is to show the user that he can use the frame variable and depending on that he can do some logic to decide whether to replace the value or give the error. Since, this sample shows that exactly, IMO this sample is enough without complications. WDYT?

MaryamZi: But why are we arbitrarily checking for John? That's my main concern. We're saying we're trying to make it more like a real-world example, but we aren't entirely.

rdulmina: The idea is there, right? Shall we change the name to `admin` as you suggested? We can change the logic to something more natural but it will complicate the logic. IMO even though the sample is not entirely a real-world case, it demonstrates the real-world purpose or the idea without complications. WDYT?

examples/on-conflict-clause/on_conflict_clause.bal
```diff
@@ -7,32 +7,53 @@ type Student record {|
...
-            {id: 2, name: "Jane", score: 150},
-            {id: 1, name: "John", score: 200}
+            {id: 1, name: "Admin", score: 100},
+            {id: 2, name: "John", score: 150},
+            {id: 1, name: "Admin", score: 200}
         ];
 
-    // The result of the following will be an error since the key `John` is duplicated.
+    // The result of the following will be an error since the key `Admin` is duplicated
+    // and `isSafeReplace()` function returns an error for the name `Admin`.
```
MaryamZi, line 17:
```suggestion
    // and the `isSafeReplace()` function returns an error for the name `Admin`.
```

MaryamZi: key in the previous line vs name here could be confusing.

examples/on-conflict-clause/on_conflict_clause.bal
```diff
@@ -7,32 +7,53 @@ type Student record {|
 |};
 
 public function main() {
-    error onConflictError = error("Key Conflict", message = "record with same key exists.");
-
     table<Student> students = table [
-            {id: 1, name: "John", score: 100},
-            {id: 2, name: "Jane", score: 150},
-            {id: 1, name: "John", score: 200}
+            {id: 1, name: "Admin", score: 100},
```
MaryamZi, line 11: Why would there be an admin in students? We need to rethink the entire example if we are changing it like this.

examples/on-conflict-clause/on_conflict_clause.bal
```diff
@@ -7,32 +7,53 @@ type Student record {|
...
 
-    // The value `100` of the key `John` will be replaced by the second occurrence value `200`.
-    map<int> lastRoundScore = map from var student in students
-                              select [student.name, student.score]
-                              on conflict ();
-
-    io:println(lastRoundScore);
-
-    // The result of the following will be an error since the key `1` is duplicated.
+    // The result of the following will be an error since the key `1` is duplicated
```
MaryamZi, line 24: key 1?

examples/on-conflict-clause/on_conflict_clause.bal
```diff
@@ -7,32 +7,53 @@ type Student record {|
...
     io:println(studentScoresTable);
+
+    table<Student> students2 = table [
+            {id: 1, name: "Admin", score: 100},
+            {id: 2, name: "Mike", score: 150},
+            {id: 2, name: "Mike", score: 200}
+        ];
+
+    // The value `100` of the key `Mike` will be replaced by the second occurrence value `200`
+    // since `isSafeReplace()` function returns `null` for the name 'Mike`.
```
MaryamZi, line 39: `null` -> nil

null is meant for JSON compatibility only.

examples/on-conflict-clause/on_conflict_clause.bal
```diff
@@ -7,32 +7,53 @@ type Student record {|
...
+    map<int>|error lastRoundScore = map from var student in students2
+                                    select [student.name, student.score]
+                                    on conflict isSafeReplace(student.name);
+
+    io:println(lastRoundScore);
+
+    // The return value will be just the construct type if the `on conflict` clause
+    // always returns `()`.
```
MaryamZi, line 47: null/nil vs () is inconsistent.

examples/on-conflict-clause/on_conflict_clause.bal
```diff
@@ -7,32 +7,53 @@ type Student record {|
...
+    map<int> lastRoundScore2 = map from var student in students2
+                               select [student.name, student.score]
+                               on conflict ();
+
+    io:println(lastRoundScore2);
+}
+
+function isSafeReplace(string name) returns error? {
```
MaryamZi, line 55: This sounds like a function that would return a boolean, but isn't.

```suggestion
function replaceIfAllowed(string name) returns error? {
```

or

```suggestion
function replaceIfSafe(string name) returns error? {
```

examples/on-conflict-clause/on_conflict_clause.md
MaryamZi: null part is still not resolved.

examples/on-conflict-clause/on_conflict_clause.bal
rdulmina: Changed the record name to `Employee`

examples/on-conflict-clause/on_conflict_clause.bal
rdulmina: Yes in this case the `id` is the key, not the `name`, and the value `1` for the `id` is duplicated

examples/on-conflict-clause/on_conflict_clause.bal
```diff
@@ -1,38 +1,61 @@
 import ballerina/io;
 
-type Student record {|
+type Employee record {|
     readonly int id;
     string name;
+    string role;
     int score;
```
MaryamZi, line 7: Why score instead of something like salary?

examples/on-conflict-clause/on_conflict_clause.bal
MaryamZi: I'm still not convinced this is the correct example. We're assuming the first entry in the table is the correct entry. 

@gimantha any thoughts on this? Since we're going back and forth. :)

examples/on-conflict-clause/on_conflict_clause.bal
```diff
@@ -1,38 +1,61 @@
...
+    table<Employee> employees = table [
+            {id: 1, name: "Mike", role: "Admin", score: 100},
+            {id: 2, name: "John", role: "HR", score: 150},
+            {id: 3, name: "Amy", role: "HR", score: 150},
+            {id: 1, name: "Mike", role: "Admin", score: 200}
         ];
 
-    // The result of the following will be an error since the key `John` is duplicated.
-    map<int>|error studentScores = map from var {name, score} in students
+    // The result of the following will be an error since the key `Mike` is duplicated
```
MaryamZi, line 18: We should try and avoid the term `key` here since we're working with tables and this isn't really a key.

examples/on-conflict-clause/on_conflict_clause.bal
```diff
@@ -1,38 +1,61 @@
...
+                                   on conflict replaceIfSafe(role);
+
+    io:println(employeeScores);
 
-    io:println(studentScores);
+    // The result of the following will be an error since the key `1` is duplicated
+    // and `replaceIfSafe()` function returns an error for the role `Admin`.
+    table<Employee> key(id)|error employeeScoresTable = table key(id) from var employee in employees
+                                                      select employee
+                                                      on conflict replaceIfSafe(employee.role);
```
MaryamZi, line 30: Here again, it's a bit confusing to replace by checking role when the id is where the conflict is, IMO.

examples/on-conflict-clause/on_conflict_clause.md
MaryamZi: Still not resolved.

examples/on-conflict-clause/on_conflict_clause.bal
rdulmina: +1 for salary

examples/on-conflict-clause/on_conflict_clause.bal
rdulmina: That is the idea right if there is a conflicting key we make decisions using the frame information. Not only the key

examples/index.json
```diff
@@ -905,6 +905,13 @@
         "verifyOutput": true,
         "isLearnByExample": true
       },
+      {
+        "name": "On conflict policy",
```
gimantha, line 909: lets remove the `policy` word here.

gimantha: we need to remove all occurrences of `policy`

rdulmina: Since we already have`BBE` named `On conflict clause` we can't just remove the `policy` word. We need to think of a better name If `On conflict policy` is not the best match

gimantha (approved):

PR #6083, opened 2025-06-04
Improve `Readonly Objects and Classes` section to avoid confusion for `readonly` and `final`

dulajdilshan:
## Purpose
$title

When reading this section, IMO, it would be better to not to have `final` keyword in the field. (Fields with `final` keyword for non-readonly classes also cannot be modified)

Ref: https://ballerina.io/learn/by-example/readonly-objects-and-classes/

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests

## Security checks

### DIFF
--- a/examples/readonly-objects-and-classes/readonly_objects_and_classes.bal
+++ b/examples/readonly-objects-and-classes/readonly_objects_and_classes.bal
@@ -9,9 +9,9 @@ readonly class FixedTimeZone {
     // Include the `readonly` object type named `TypeZone` using object type inclusion.
    *TimeZone;
 
-   // The `final decimal` field named offset is read-only because the `decimal` basic type is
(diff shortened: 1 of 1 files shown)


examples/readonly-objects-and-classes/readonly_objects_and_classes.bal
```diff
@@ -9,9 +9,9 @@ readonly class FixedTimeZone {
     // Include the `readonly` object type named `TypeZone` using object type inclusion.
    *TimeZone;
 
-   // The `final decimal` field named offset is read-only because the `decimal` basic type is
+   // The `decimal` field named offset is read-only because the `decimal` basic type is
```
MaryamZi, line 12: I think this needs to include that the class is readonly.

PR #6131, opened 2025-08-25
[2201.12.x] Add messaging package BBEs

TharmiganK:
## Purpose

> $Subject

Part of https://github.com/ballerina-platform/ballerina-library/issues/8193


examples/message-store-listener/message_store_listener.md
```diff
@@ -0,0 +1,16 @@
+# Message Store Listener
+
+The `messaging` library provides a message store listener which polls the store at a configurable interval, dispatches messages to an attached service, and manages retries and Dead-Letter failed messages.
```
TharmiganK:
```suggestion
The `messaging` library provides a message store listener which polls the store at a configurable interval, dispatches messages to an attached service, and manages retries and a Dead-Letter store.
```

examples/index.json
```diff
@@ -4473,6 +4473,28 @@
...
+        "name": "Message store",
+        "url": "message-store",
+        "verifyBuild": true,
+        "verifyOutput": false,
+        "disablePlayground": false,
+        "isLearnByExample": true
+      },
+      {
+        "name": "Message Store Listener",
+        "url": "messages-store-listener",
```
TharmiganK:
```suggestion
        "url": "message-store-listener",
```

examples/message-store/message_store.bal
```diff
@@ -0,0 +1,42 @@
+import ballerina/messaging;
+import ballerina/io;
+
+public isolated client class CustomStore {
```
ThisaruGuruge: Is there a purpose of having this code here? We are not using this CustomStore type in the example, right? Maybe if we can show a simple implementation here? Maybe using a file? Will that make the example complex?

TharmiganK: Wanted to emphasise the fact that we can implement the type to create custom message store.

Using a file based message store can complicate the example. I'll refactor this to have a custom store implementation as a separate block

TharmiganK: I have split this to two source files in [923ae5b](https://github.com/ballerina-platform/ballerina-distribution/pull/6131/commits/923ae5b7dfdbdd73fda0034878f87388cc4f6106)

But the custom store does not have an implementation there. Having an implementation may complicate this BBE, since the design should incorporate a message order and it should support isolation

ThisaruGuruge: In that case, maybe we can introduce a new BBE to showcase the custom store implementation? I think having the unimplemented custom store makes the BBE complex. Another point is that we usually expect users to run BBEs by copy and pasting. This may not be the case here. 
@anupama-pathirage @MaryamZi thoughts?

MaryamZi: I also prefer having at least a simple (in-memory) implementation.

TharmiganK: Ack

TharmiganK: Added a new BBE for implementing the custom message store with [d622483](https://github.com/ballerina-platform/ballerina-distribution/pull/6131/commits/d62248393f66009bbf9c60c9da72749d2c0f650c)

PR #6147, opened 2025-09-12
Remove OpenAPI dependencies from the distribution

Nuvindu:

### DIFF
--- a/ballerina-test/build.gradle
+++ b/ballerina-test/build.gradle
@@ -54,6 +54,7 @@ task distributionTest {
         systemProperty "spec.version", "$specVersion"
         systemProperty "tool.version", "$ballerinaCommandVersion"
         systemProperty "code.name", "$project.codeName"
+        systemProperty "openapi.version", "$openapiModuleVersion".split("-")[0]
 
         useTestNG() {
             suites 'src/test/resources/testng.xml'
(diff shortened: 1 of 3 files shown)


gradle.properties
```diff
@@ -91,8 +89,8 @@ graphqlToolVersion=0.14.0-20250523-094200-b95690f
 # Protoc Tool
 protocToolVersion=1.0.0-20250528-120700-b5350a0
 
-# OpenAPI Module
-openapiToolVersion=2.3.2
+# OpenAPI Tool
+openapiToolVersion=2.4.0
```
TharmiganK: Do we use the same version to include `ballerina/openapi` package in the distribution?

TharmiganK: We should use a timestamped version here right?

Nuvindu: Fixed by [d79ba88](https://github.com/ballerina-platform/ballerina-distribution/pull/6147/commits/d79ba889e775feff5cc71658eaf563de48675ec5)

Nuvindu: Used two separate versions for each dependency (`tool.openapi`, `openapi`)

Fixed by [d0ee28f](https://github.com/ballerina-platform/ballerina-distribution/pull/6147/commits/d0ee28f70af6b982e2564f62d9d30beabac3d2a2)

PR #6425, opened 2026-08-19
[2201.13.x] Add Ballerina By Example samples for the zip library

niveathika:
## Purpose

Adds Ballerina By Example samples for the `zip` library, covering the convenience APIs only — `compress`, `listEntries`, and `decompress`. The `ArchiveWriter`/`ArchiveReader` entry-by-entry APIs are left out.

The samples land in a new **Zip** group under **Common libraries**, placed right after UUID in column 3.

| Sample | Shows |
| --- | --- |
| Create a ZIP archive (`zip-compress`) | `compress` with `CompressOptions` — compression level, `overwrite`, and how `includeSourceDirectory` names the entries |
| List archive entries (`zip-list-entries`) | `listEntries` returning `zip:Entry` metadata without unpacking the archive |
| Extract a ZIP archive (`zip-decompress`) | `decompress` into a directory, with `fileWriteMode` deciding what happens to a file already there |
| Unpack an untrusted archive (`zip-decompress-untrusted`) | `ExtractionLimits` capping what an extraction may cost, and telling refusals apart by `LimitExceededError` / `UnsafePathError` |

Each sample creates its own fixture files, so it runs standalone with no prerequisites.

The index entries carry `needCentral: true`, since `zip` is not packed with the distribution and resolves from Central (`ballerina/zip:0.9.1`).

## Checklist
- [x] Every sample was run with `bal run` on Ballerina 2201.13.3, and the `.out` files hold that real output
- [x] All `.bal` files are `bal format` clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)


examples/zip-compress/zip_compress.bal
```diff
@@ -0,0 +1,21 @@
...
+import ballerina/zip;
+
+public function main() returns error? {
+    // Create the directory of reports to archive.
+    check file:createDir("reports");
+    check io:fileWriteString("reports/region-totals.csv",
+            "region,quarter,total\nEMEA,Q1,4820\nAPAC,Q1,3915\nAMER,Q1,5210\n");
+    check io:fileWriteString("reports/notes.txt",
+            "Compiled for the August review. The regional totals are " +
+            "provisional until the audit closes.\n");
```
niveathika: This will be prerequiste. we give a resources/reports dir with all files

examples/zip-decompress-untrusted/zip_decompress_untrusted.bal
```diff
@@ -0,0 +1,36 @@
...
+import ballerina/io;
+import ballerina/zip;
+
+public function main() returns error? {
+    // Stands in for an archive that arrived from outside the system.
+    check file:createDir("uploads");
+    check io:fileWriteString("uploads/q1.csv", "region,total\nEMEA,4820\n");
+    check io:fileWriteString("uploads/q2.csv", "region,total\nEMEA,5310\n");
+    check io:fileWriteString("uploads/q3.csv", "region,total\nEMEA,3915\n");
+    check zip:compress("uploads", "uploaded.zip", {overwrite: true});
```
niveathika: Same

examples/zip-decompress/zip_decompress.bal
```diff
@@ -0,0 +1,20 @@
+import ballerina/file;
+import ballerina/io;
+import ballerina/zip;
+
+public function main() returns error? {
+    check file:createDir("reports");
+    check io:fileWriteString("reports/notes.txt",
+            "The regional totals are provisional until the audit closes.");
+    check zip:compress("reports", "reports.zip", {overwrite: true});
```
niveathika: Same

examples/zip-list-entries/zip_list_entries.bal
```diff
@@ -0,0 +1,23 @@
...
+import ballerina/zip;
+
+public function main() returns error? {
+    check file:createDir("reports");
+    check io:fileWriteString("reports/region-totals.csv",
+            "region,quarter,total\nEMEA,Q1,4820\nAPAC,Q1,3915\nAMER,Q1,5210\n");
+    check io:fileWriteString("reports/notes.txt",
+            "Compiled for the August review. The regional totals are " +
+            "provisional until the audit closes.\n");
+    check zip:compress("reports", "reports.zip", {overwrite: true});
```
niveathika: Same

examples/index.json
```diff
@@ -5405,6 +5405,49 @@
...
+      },
+      {
+        "name": "Extract a ZIP archive",
+        "url": "zip-decompress",
+        "verifyBuild": true,
+        "verifyOutput": false,
+        "disablePlayground": true,
+        "isLearnByExample": false,
+        "needCentral": true
+      },
```
niveathika: Maybe we can combine this? First read entries. Decompress and read a file?

examples/zip-compress/zip_compress.bal
niveathika: [Claude Code] Done in 0373595. The sample no longer creates anything — it ships `resources/reports/` with `region-totals.csv` and `notes.txt`, compresses `resources/reports`, and the `.md` carries a `## Prerequisites` section naming that directory.

examples/zip-list-entries/zip_list_entries.bal
niveathika: [Claude Code] Done in 0373595 — this sample is gone, folded into `zip-decompress` per your `index.json` comment. The merged sample takes `reports.zip` from the Create a ZIP archive example as its prerequisite instead of writing files.

examples/zip-decompress/zip_decompress.bal
niveathika: [Claude Code] Done in 0373595. No fixture code left; the prerequisite is running the Create a ZIP archive example to put `reports.zip` in the current directory, stated in the `.md`.

examples/zip-decompress-untrusted/zip_decompress_untrusted.bal
niveathika: [Claude Code] Done in 0373595. Same here — no fixture code; it unpacks `reports.zip` from the Create a ZIP archive example, whose three entries trip `maxEntries: 2` and produce the recorded `LimitExceededError` output.

examples/index.json
niveathika: [Claude Code] Done in 0373595. `zip-list-entries` is deleted and `zip-decompress` now does all three steps: `listEntries`, then `decompress`, then reading `notes.txt` back out of the unpacked directory. The Zip group is down to three samples — Create / Extract / Unpack an untrusted archive — and each entry now carries `disableVerificationReason: "Needs prerequisite condition"`.

examples/zip-decompress-untrusted/zip_decompress_untrusted.bal
```diff
@@ -0,0 +1,28 @@
...
+
+    // The error type tells a hostile archive from a broken one. An entry whose
+    // name would write outside the target directory is refused in the same way,
+    // with an `UnsafePathError`.
+    if result is zip:LimitExceededError {
+        io:println("Refused: ", result.message());
+    } else if result is zip:UnsafePathError {
+        io:println("Refused: an entry writes outside the target directory");
+    } else {
+        check result;
```
Nuvindu: Do we need this line?

Nuvindu: Can we try something like this?

```ballerina
if result is zip:LimitExceededError {
    io:println("Refused: ", result.message());
} else if result is zip:UnsafePathError {
    io:println("Refused: an entry writes outside the target directory");
} else if result is zip:Error {
    io:println("Refused: failed to decompress the archive: ", result.message());
} else {
    io:println("Unpacked reports.zip");
}
```

examples/zip-decompress/zip_decompress.bal
```diff
@@ -0,0 +1,25 @@
...
+
+public function main() returns error? {
+    // Read what the archive holds without unpacking it. The entries come back
+    // in the order they are stored, directories included, each carrying the
+    // sizes, the compression method, the modified time, and the CRC-32 of the
+    // content.
+    zip:Entry[] entries = check zip:listEntries("reports.zip");
+    foreach zip:Entry entry in entries {
+        io:println(entry.name, ": ", entry.uncompressedSize, " -> ",
+                entry.compressedSize, " bytes (", entry.method, ")");
```
Nuvindu: Can we use string templates for this?

```suggestion
        io:println(string `${entry.name}: ${entry.uncompressedSize} -> ${entry.compressedSize} bytes (${entry.method})`);
```

examples/zip-decompress-untrusted/zip_decompress_untrusted.bal
niveathika: [Claude Code] No, it isn't needed — taken as suggested in 11b49f8. `check result` is replaced by an explicit `else if result is zip:Error` branch, so every outcome is spelled out in the chain, which is the point of the sample. Nothing propagates any more, so `main` also drops `returns error?`.

examples/zip-decompress/zip_decompress.bal
niveathika: [Claude Code] Switched to string templates in 11b49f8, split across two prints. BBE lines have to stay under 80 characters — the code renders in a fixed-width box on the by-example page, so a 118-character line scrolls sideways — and the single-expression template does not fit. `io:print` for the first half keeps each entry on one output line, so the recorded output is unchanged:

```ballerina
io:print(string `${entry.name}: ${entry.uncompressedSize} -> `);
io:println(string `${entry.compressedSize} bytes (${entry.method})`);
```

Nuvindu (approved):
LGTM

PR #6430, opened 2026-08-20
Add BBEs for the EDI envelope API

niveathika:
## Purpose

Part of the EDI documentation pass for the envelope-aware API ([ballerina-spec#1441](https://github.com/ballerina-platform/ballerina-spec/issues/1441)).

The EDI group has two examples, `edi-to-record` and `record-to-edi`. Both predate the envelope-aware API released in `edi` 1.6.0, and both carry `verifyBuild: false` / `verifyOutput: false` because they depend on generated code that is not in the repository. So nothing in Ballerina by Example shows schema-free header reading or fail-safe interchange parsing, and nothing in the EDI group is verified.

## Changes

Two new BBEs, each with its schema and sample interchange inline, so they compile, run, and have their **output verified** (`verifyBuild: true`, `verifyOutput: true`):

- **Read EDI envelope headers** (`edi-envelope-headers`) — `x12HeadersFromEdiString` and `edifactHeadersFromEdiString`. No schema: the ISA/GS and UNB/UNH layouts are built in, which is the fastest way to identify a trading partner before deciding how to parse the rest.
- **Parse an EDI interchange** (`edi-interchange-parsing`) — `interchangeFromEdiString` against a schema declaring an `envelope`, showing that envelope segments are fail-fast while transaction bodies are fail-safe. The bundled interchange holds two messages, the second missing its mandatory `BGM`, so the quarantine path is exercised.

Both verified locally with `bal run <file>.bal`; the output matches the `.out` files exactly.

### Switched to the published EDIFACT packages

Now that `ballerinax/edifact.d03a.*` is republished with the envelope-aware API, three of the four BBEs import it instead of carrying a schema:

- **EDI to record conversion** and **Record to EDI conversion** were `verifyBuild: false, verifyOutput: false` with `disableVerificationReason: "Includes generated code"` — they imported a generated submodule that was never in the BBE directory, so neither could ever be built. Both now call `mORDERS:fromEdiString` / `mORDERS:toEdiString` on `ballerinax/edifact.d03a.supplychain`, which makes them **verifiable for the first time** (`verifyBuild: true, verifyOutput: true`). That also removed the schema files and the eight `.out` files of project/codegen ceremony; the two now form a round trip, since the writer's output is exactly the reader's input.
- **Parse an EDI interchange** drops its inline schema for `mORDERS:interchangeFromEdiString`, keeping the fail-safe body demonstration.
- **Read EDI envelope headers** is unchanged: it uses the schema-free `x12HeadersFromEdiString` / `edifactHeadersFromEdiString`, so no message module is involved.

`needCentral: true` on the three that import `ballerinax`. Worth noting for review: no existing `needCentral` BBE sets `verifyOutput: true`, but that is a coincidence of which samples happen to need Central, not a constraint — `needCentral` only drops `--sticky --offline` from `bal build` (`ballerina/build.gradle:964`), and `outputVerification` runs `bal run <file>.bal` independently. Verified that a single-file `bal run` resolves the `ballerinax` import from Central, which is exactly what output verification does.

All four verified locally; the `.out` files match byte for byte.

### Line items are deliberately absent

The reader BBE uses `BGM`, `DTM`, `RFF`, and `NAD` rather than `LIN`/`QTY`, because the published D03A schemas drop the `LIN` element `1229` (Action request/notification code). With that element missing the fields shift, so the conventional `LIN+1++A-250:IN'` silently lands the item identifier in `SUB_LINE_INFORMATION` with no error. Rather than print a wire format no partner sends, the BBE stays on segments the converter handles correctly (`BGM`, `DTM`, `RFF`, `NAD`, all checked against the D03A EDSD).

Root cause, for whoever picks it up: in the D03A segment directory the name of element `1229` wraps onto a continuation line, and `convertEdifactSchema` drops elements whose name wraps. 17 elements across 17 segments in the D03A directory are affected; of the segments that appear in ORDERS, `LIN`, `DGS`, `PCI`, and `RCS` each lose exactly the wrapped element and keep every element after it. This is in the source schemas in `ballerina-platform/edifact` as well, so it is a converter issue rather than a packaging one.


examples/edi-to-record/edi_to_record.bal
```diff
@@ -1,12 +1,35 @@
 import ballerina/io;
-import edi_to_record.sorder;
+
```
niveathika:
```suggestion
```

examples/edi-to-record/edi_to_record.md
```diff
@@ -1,37 +1,15 @@
...
+Envelope segments are skipped, so `fromEdiString` accepts a bare message as well as a full interchange. Segments that repeat, such as `RFF` and `NAD`, become segment groups on the record.
 
 ::: code edi_to_record.bal :::
 
-Run the program using the command below:
+Run the program using the command below.
 
 ::: out output.out :::
+
+When a trading partner deviates from the published specification — an extra element, a different code list, a segment the standard marks optional but the partner always sends — convert that specification into a Ballerina EDI schema, edit the schema, and generate a module from it with the `bal edi` tool instead of using the prebuilt package.
```
niveathika: Link to learn page which contains the steps

examples/record-to-edi/record_to_edi.bal
```diff
@@ -1,12 +1,37 @@
 import ballerina/io;
-import record_to_edi.sorder;
+
```
niveathika:
```suggestion
```

examples/edi-to-record/edi_to_record.bal
niveathika: [Claude Code] Removed in 1a5af30. Checked the convention first — every other BBE that imports both (`aws-lambda-s3-trigger`, `azure-functions-hello-world`, `cdc-advanced-service`) keeps `ballerina/` and `ballerinax/` adjacent with no separator. Re-ran the BBE; output unchanged.

examples/edi-to-record/edi_to_record.md
niveathika: [Claude Code] Linked in 1a5af30 to `/learn/edi-tool/#changing-the-specification-for-a-trading-partner`, the new subsection carrying the convert → edit → regenerate steps.

Note the anchor only resolves once ballerina-platform/ballerina-dev-website#10395 merges, since that section is added there.

examples/record-to-edi/record_to_edi.bal
niveathika: [Claude Code] Same fix in 1a5af30; output re-verified.

examples/edi-to-record/edi_to_record.md
```diff
@@ -1,37 +1,15 @@
...
 
 ::: code edi_to_record.bal :::
 
-Run the program using the command below:
+Run the program using the command below.
+
+::: out edi_to_record.out :::
 
-::: out output.out :::
+When a trading partner deviates from the published specification — an extra element, a different code list, a segment the standard marks optional but the partner always sends — convert that specification into a Ballerina EDI schema, edit the schema, and generate a module from it with the `bal edi` tool instead of using the prebuilt package. See [Changing the specification for a trading partner](https://ballerina.io/learn/edi-tool/#changing-the-specification-for-a-trading-partner) for the steps.
```
YasanPunch: This anchor doesn't seem to exist, the EDI tool page has no "Changing the specification for a trading partner" section?

examples/edi-interchange-parsing/edi_interchange_parsing.metatags
```diff
@@ -0,0 +1,2 @@
+description: This BBE demonstrates parsing a full EDI interchange into its envelope hierarchy with fail-safe transaction bodies, using the Ballerina `edi` library.
```
YasanPunch: here it says "using the Ballerina edi library", but this example only imports ballerinax/edifact.d03a.supplychain, right?

examples/edi-to-record/edi_to_record.md
niveathika: This is the custom schema as discussused in edi-tool.md. Maybe better to use uniform language.

examples/edi-interchange-parsing/edi_interchange_parsing.metatags
niveathika: [Claude Code] Right — that example imports only `ballerinax/edifact.d03a.supplychain.mORDERS`. Reworded to "using a prebuilt EDIFACT package" in 85c25d1. `edi-envelope-headers` does use `ballerina/edi` directly, so its description is unchanged.

examples/edi-to-record/edi_to_record.md
niveathika: [Claude Code] The section exists on ballerina-platform/ballerina-dev-website#10395 but not yet on the published page, which is why the anchor 404s today.

On the wording: there were three names for one concept — "Changing the specification" (tool page), "Adjusting a schema" (integration docs), "Adapt the schema" (guide). All four surfaces now read **"Adapting the schema to a trading partner"**: c15061b (tool page heading and anchor), 85c25d1 (this link), wso2/docs-integrator#633, and wso2/docs-integrator#631. The specification itself never changes — the schema derived from it does — so the old title was the misleading one.

Nuvindu (approved):
LGTM

